Issue #55, opened 2022-06-22
Add `reshape` function


coreylowman:
This would only work for two tensors with the same number of elements. This can be used for #34

coreylowman:
First stab in https://github.com/coreylowman/dfdx/commit/1215ed56ed6f7ca98e5d335f35915cf4de9846a7

coreylowman:
Another approach that is slightly more readable and self contained (but still requires `#![feature(generic_const_exprs)]`):

```rust

struct True;

trait ConstAssert {
    type Result;
}

struct ConstEq<const A: usize, const B: usize>;

impl<const A: usize> ConstAssert for ConstEq<A, A> {
    type Result = True;
}

fn copy<Lhs: Count, Rhs: Count>(lhs: &Lhs, rhs: &mut Rhs)
where
    ConstEq<{ Lhs::NUM_ELEMENTS }, { Rhs::NUM_ELEMENTS }>: ConstAssert<Result = True>,
{
    let l = lhs.ref_first() as *const f32;
    let r = rhs.mut_first() as *mut f32;
    unsafe {
        std::ptr::copy_nonoverlapping(l, r, Lhs::NUM_ELEMENTS);
    }
}
```

PR #63, opened 2022-07-08
Draft: #55 Initial implementation of reshape

coreylowman:


src/tensor_ops/reshape.rs
```diff
@@ -0,0 +1,68 @@
...
+
+pub struct True;
+pub trait ConstAssert {
+    type Result;
+}
+
+pub struct ConstEq<const A: usize, const B: usize>;
+impl<const N: usize> ConstAssert for ConstEq<N, N> {
+    type Result = True;
+}
```
aidanalphafund, line 41: I think a slightly cleaner set of traits would be the following:
```rust
pub trait ConstTrue{}
pub struct ConstEq<const A: usize, const B: usize>;

impl<const N: usize> ConstTrue for ConstEq<N, N> {}
```

This would change your where clause above from this:
```rust
ConstEq<{ T::Array::NUM_ELEMENTS }, { R::Array::NUM_ELEMENTS }>: ConstAssert<Result = True>
```

To this:
```rust
ConstEq<{ T::Array::NUM_ELEMENTS }, { R::Array::NUM_ELEMENTS }>: ConstTrue
```

coreylowman: Oooh this is great thanks!

Issue #66, opened 2022-07-10
Support MacOS with Intel MKL


coreylowman:
MacOS support needs `REDIST_DIRS`, `LINK_DIRS`, and `LINK_LIBS`.

_Originally posted by @coreylowman in https://github.com/coreylowman/dfdx/pull/64#discussion_r917408808_

coreylowman:
Root issue #18

yerke:
Oops. I didn't that there was a separate issue for macOS and opened https://github.com/coreylowman/dfdx/issues/72.

I was able to run all examples (except for mnist since I don't have the input files) successfully on macOS 12.4 with MBP 16 M1 Max.

I wasn't able to install Intel oneAPI Base Toolkit. It breaks during installation on Distribution for Python component with the error below. Do I need the whole Toolkit or only parts of it?

```
07/13/2022 17:51:04:545 : 247755 : ERROR : Installation of component has failed.
Component id: intel.oneapi.mac.python3, name: Intel® Distribution for Python*, version: 2022.1.0-213.
During the execution of the application '/var/intel/installercache/downloadcache/intel.oneapi.mac.python3,v=2022.1.0-213/exec_script.sh' with arguments '/opt/intel/oneapi' errors were received:'+ xargs /bin/sh
'
Error: Sequence execution failed.
```

joseph-x-li:
I doubt we can install Intel MKL on M1 macs. I have access to an Intel Mac, will get back with an attempt on that one. By the way, I am "YRUDancin" from reddit.

coreylowman:
@yerke you only need one part of the install. i think it's called something like "Intel oneAPI Math Kernel Library"? You shouldn't need to install anything for python.

coreylowman:
> I doubt we can install Intel MKL on M1 macs. I have access to an Intel Mac, will get back with an attempt on that one. By the way, I am "YRUDancin" from reddit.

Can you expand on this? I had assumed since the [mkl install page](https://www.intel.com/content/www/us/en/developer/tools/oneapi/base-toolkit-download.html?operatingsystem=mac&distributions=webdownload) had macos as one of the dropdown options that it would be possible.

joseph-x-li:
I assume Intel MKL uses the x86-64 ISA, so it will only work with Macs with an Intel chip. M1 macs are all arm64e. You will likely have to fall back to a different BLAS implementation for M1 Macs. 

Anyways, I have MKL and `mkl_link_tool`, what command would you like me to run?

[...]

PR #73, opened 2022-07-14
add macOS support

yerke:
Closes https://github.com/coreylowman/dfdx/issues/66

Tested on macOS 12.4 with MBP 14 Intel with all features (`mkl-static-iomp`, `mkl-static-seq`, `mkl-dynamic-iomp`, `mkl-dynamic-seq`) and without any of them.

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,0 +1,2 @@
+[target.x86_64-apple-darwin]
+rustflags = ["-C", "link-arg=-Wl,-rpath,/opt/intel/oneapi/compiler/latest/mac/compiler/lib"]
--- a/build.rs
+++ b/build.rs
@@ -24,7 +24,7 @@
 //! - [x] Linux 32 bit
 //! - [x] Linux 64 bit
 //! - [ ] MacOS 32 bit
-//! - [ ] MacOS 64 bit
+//! - [x] MacOS 64 bit
 
 pub const MKL_VERSION: &str = "2022.1.0";
 pub const STATIC: bool = cfg!(feature = "mkl-static-seq") || cfg!(feature = "mkl-static-iomp");
@@ -37,8 +37,12 @@ pub const LINK_TYPE: &str = if STATIC { "static" } else { "dylib" };
 pub const LIB_POSTFIX: &str = if cfg!(windows) && DYNAMIC { "_dll" } else { "" };
 pub const LD_DIR: &str = if cfg!(windows) {
     "PATH"
-} else {
+} else if cfg!(target_os = "linux") {
     "LD_LIBRARY_PATH"
+} else if cfg!(target_os = "macos") {
+    "DYLD_LIBRARY_PATH"
+} else {
+    ""
 };
 
 pub const DEFAULT_ONEAPI_ROOT: &str = if cfg!(windows) {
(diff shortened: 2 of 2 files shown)


build.rs
```diff
@@ -24,7 +24,7 @@
 //! - [x] Linux 32 bit
 //! - [x] Linux 64 bit
 //! - [ ] MacOS 32 bit
```
yerke, line 26: I don't have a 32 bit Apple machine, and so I have no way to test it.

.cargo/config
```diff
@@ -0,0 +1,2 @@
+[target.x86_64-apple-darwin]
+rustflags = ["-C", "link-arg=-Wl,-rpath,/opt/intel/oneapi/compiler/latest/mac/compiler/lib"]
```
yerke, line 2: Work around rpath issue: https://github.com/coreylowman/dfdx/issues/66#issuecomment-1183927838

build.rs
joseph-x-li: Last year Apple made 32-bit machines was 2011, so I think it is safe to remove this line.
```suggestion

```

build.rs
```diff
@@ -102,6 +105,15 @@ pub const LINK_DIRS: &[&str] = &[
...
 pub const SHARED_LIB_DIRS: &[&str] = LINK_DIRS;
 
+#[cfg(target_os = "macos")]
+const MACOS_COMPILER_PATH: &str = "compiler/latest/mac/compiler/lib";
+
+#[cfg(target_os = "macos")]
+pub const LINK_DIRS: &[&str] = &[MACOS_COMPILER_PATH, "mkl/latest/lib"];
+
+#[cfg(target_os = "macos")]
+pub const SHARED_LIB_DIRS: &[&str] = LINK_DIRS;
```
coreylowman, line 115: I see MACOS_COMPILER_PATH is explicitly skipped in the shared lib dirs check below. Let's just remove it from shared lib dirs so we don't need that explicit skip. SHARED_LIB_DIRS is only used to the path checking, so it isn't actually passed to rustc at all.

```suggestion
#[cfg(target_os = "macos")]
pub const LINK_DIRS: &[&str] = &["compiler/latest/mac/compiler/lib", "mkl/latest/lib"];

#[cfg(target_os = "macos")]
pub const SHARED_LIB_DIRS: &[&str] = &["mkl/latest/lib"];
```

build.rs
```diff
@@ -176,3 +203,17 @@ fn main() -> Result<(), BuildError> {
...
     Ok(())
 }
+
+#[cfg(any(
+    feature = "mkl-static-iomp",
+    feature = "mkl-static-seq",
+    feature = "mkl-dynamic-iomp",
+    feature = "mkl-dynamic-seq"
+))]
+fn suggest_setvars_cmd(root: &str) -> String {
```
coreylowman, line 213: Great clean up! I think we can just remove this outer cfg block, nothing wrong with having it sit around for all builds.
```suggestion
fn suggest_setvars_cmd(root: &str) -> String {
```

coreylowman (changes requested):
I think this looks great. Would you be able to test out if the `cargo:rustc-link-arg` works in place of the .cargo/config file? Let me know if you plan on spending more time on this, otherwise I can take this over. Really want to get this merged in!

yerke:
Sorry @coreylowman, I got too busy with my work. Thanks for finishing it.

coreylowman:
No worries at all thanks for the initial PR @yerke!!

PR #83, opened 2022-07-14
Added custom model example

jafioti:
I think it's important to show how to build a custom module that implements custom functionality so I took the mnist example and created a custom MLP struct. Obviously this is a trivial scenario, but it demonstrates how to use the traits to implement a new module for both single and batch forward functions.


coreylowman (changes requested):
Yeah I definitely want to merge this in, but let's reduce it a bit to make it as clear as possible! Let me know if you aren't planning on updating and I can take this over.

jafioti:
Alright I simplified the example to just a single forward and batched forward. Let me know what you think.

coreylowman (approved):
Looks great! Thanks for the contribution 🎉

PR #85, opened 2022-07-17
add Debug and Display support for `NpzError`

XBagon:


coreylowman (approved):
Thanks for this contribution!

Issue #87, opened 2022-07-17
Add feature when compiling in nightly & include generic_const_exprs if that feature is enabled


coreylowman:
> I'd imagine you could do something like use cfg tags to check if they're on nightly. Something like this: https://stackoverflow.com/questions/49247618/check-whether-compiler-is-nightly-at-compile-time

> that way no other features need to be enabled, it just works on nightly.

_Originally posted by @jafioti in https://github.com/coreylowman/dfdx/issues/43#issuecomment-1186588385_

PR #90, opened 2022-07-18
#55 reshape, and #87 additional work on nightly feature

### LINKED ISSUE #55: Add `reshape` function

coreylowman:
Closes #55

### DIFF
--- a/.github/workflows/cargo-check.yml
+++ b/.github/workflows/cargo-check.yml
@@ -11,7 +11,7 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: stable
+          toolchain: nightly
           override: true
       - uses: actions-rs/cargo@v1
         with:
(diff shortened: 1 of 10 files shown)


coreylowman:
Hmm am unsure why the doctest is failing. If i put the doctest as I normal test, it passes. 🤷

coreylowman:
Quick fix: nightly features should have doctests as `no_run`. I think it's fair that nightly features not necessarily be as thorough as stable stuff. Just need to capture that somewhere

.github/workflows/cargo-check.yml
```diff
@@ -11,7 +11,7 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: stable
```
coreylowman, line 14: Revert all these

PR #94, opened 2022-07-19
Added 2d broadcast_first functions and 3d linear forward

### LINKED ISSUE #92: Batched Matrix Multiply
For transformers, it's necessary to multiply input (Tensor<Batch, Seq, Embed>) by weight (Tensor<Embed, Embed>). One way of achieving this is to flatten the input to Tensor<Batch + Seq, Embed> and then do normal multiplication, but I think there is a better way. Plus, reshaping right now isn't supported yet, and even when it is will require nightly. 

I can implement bmm, but I'm unsure what the best way to go about it is right now. `matrixmultiply` doesn't have a function for it.

jafioti:
This closes #92


jafioti:
Forgot to mention this also has a naive implementation of batch_matmul and batch_matmul_transpose. Currently just uses a sequential loop.

src/tensor_ops/arith_broadcast_outer.rs
```diff
@@ -81,6 +83,98 @@ where
...
 }
 
+/// `lhs + &rhs`. `rhs` is broadcasted `N * M` times, where `N` is the first dimension and `N` is the second dimension of `lhs`.
+///
+/// E.g If Lhs has dimension `(5, 2, 3)`, then Rhs has to be dimension `(3,)`.
+///
+/// Examples:
+/// ```rust
+/// # use dfdx::prelude::*;
+/// let a = Tensor3D::new([[[1.0, 2.0, 3.0], [4.0, 5.0, 6.0]]]);
```
coreylowman, line 93: Could you add another array at the highest dimension? Just to cover that the same array is applied to all dimensions
e.g.
```rust
let a = Tensor3D::new([
        [[1.0, 2.0, 3.0], [4.0, 5.0, 6.0]],
        [[-1.0, -2.0, -3.0], [-4.0, -5.0, -6.0]],
    ]);
```

src/tensor_ops/binary_map.rs
```diff
@@ -158,6 +158,61 @@ where
...
 }
 
+/// Apply binary function `f` to `lhs` and `rhs`, where `rhs` is broadcasted `M` times to be the same shape as `lhs`.
+/// `dfdx` and `dfdy` are the partial derivatives of f wrt. x and y respectively.
+///
+/// `f`, `dfdx`, and `dfdy` are all the same type.
+///
+/// Generics:
+/// - `N`: The first dimension of `lhs`.
+/// - `M`: The second dimension of `lhs`.
```
coreylowman, line 168: nit: Can you make `M` the first dimension? Tried to keep `M` as the outer dimension everywhere!

src/tensor_ops/matmul.rs
```diff
@@ -52,7 +52,7 @@ pub fn matmul<const M: usize, const K: usize, const N: usize, TAPE: Tape>(
 /// - `K`: number of columns of `lhs` and number of rows of `rhs`.
 /// - `N`: Number of columns of `rhs`.
 ///
-/// Returns a 2d tensor representing an MxO matrix.
+/// Returns a 2d tensor representing an MxN matrix.
```
coreylowman, line 55: 👍

src/tensor_ops/matmul.rs
```diff
@@ -159,6 +159,106 @@ pub fn vecmat_mul_transpose<const K: usize, const N: usize, TAPE: Tape>(
...
+
+    // copy rhs data for use later when computing gradients
+    let rhs_data = rhs.data.clone();
+
+    move_tape_and_add_backward_binop(lhs, rhs, result, move |lhs, rhs, result, grads| {
+        let (lhs_grad, result_grad) = grads.mut_and_ref(&lhs, &result);
+        bmm_bt(result_grad, rhs_data.as_ref(), lhs_grad);
+
+        let (rhs_grad, result_grad): (&mut [[f32; N]; K], &[[[f32; N]; M]; B]) =
+            grads.mut_and_ref(&rhs, &result);
```
coreylowman, line 199: Yeah I could not figure out why you need to type hint for these.

src/tensor_ops/matmul.rs
```diff
@@ -311,6 +411,47 @@ fn mm_atct<const M: usize, const K: usize, const N: usize>(
...
+) {
+    let b = b.as_ptr() as *const f32;
+
+    // Purely sequential for now, should parallelize using rayon or some other BLAS solution
+    for i in 0..B {
+        let a = a[i].as_ptr() as *const f32;
+        let c = c[i].as_mut_ptr() as *mut f32;
+
+        unsafe {
+            matrixmultiply::sgemm(
```
coreylowman, line 428: We'll need to add MKL support (that's the cblas calls) for `bmm` and `bmm_bt`. That can be a separate PR though, for now we can juts open an issue to track it.

src/tensor_ops/matmul.rs
```diff
@@ -159,6 +159,106 @@ pub fn vecmat_mul_transpose<const K: usize, const N: usize, TAPE: Tape>(
...
+///
+/// # Examples
+///
+/// ```rust
+/// # use dfdx::prelude::*;
+/// let x: Tensor3D<5, 3, 2> = Tensor3D::zeros();
+/// let y: Tensor2D<4, 2> = Tensor2D::zeros();
+/// let result: Tensor3D<5, 3, 4> = batch_matmul_transpose(x, &y);
+/// ```
+pub fn batch_matmul_transpose<
```
coreylowman, line 231: Should add some unit tests for  this function & `batch_matmul`. What I've been doing for unit tests is basically just call the operations in pytorch with randn lhs & rhs, running the operation, and then calling .backward() & printing out .grad. 

Not sure if there's a better way to do that!

src/tensor_ops/arith_broadcast_outer.rs
```diff
@@ -81,6 +83,98 @@ where
...
+/// let b = Tensor1D::new([-1.0, 0.0, 1.0]);
+/// let r = div_broadcast_rhs_first_2d(a, &b);
+/// assert_eq!(r.data(), &[[[-1.0, f32::INFINITY, 3.0], [-4.0, f32::INFINITY, 6.0]]]);
+/// ```
+pub fn div_broadcast_rhs_first_2d<Lhs, Rhs, const M: usize, const N: usize>(
```
coreylowman, line 167: We'll probably want to add some unit tests for these to make sure the gradients are being calculated correctly.

src/nn/linear.rs
```diff
@@ -97,6 +97,17 @@ impl<const B: usize, const I: usize, const O: usize, H: Tape> Module<Tensor2D<B,
     }
 }
 
+impl<const B: usize, const S: usize, const I: usize, const O: usize, H: Tape>
+    Module<Tensor3D<B, S, I, H>> for Linear<I, O>
+{
+    type Output = Tensor3D<B, S, O, H>;
+
+    /// Batched 3d forward using [batch_matmul()] and [add_broadcast_rhs_first()]
+    fn forward(&self, x: Tensor3D<B, S, I, H>) -> Self::Output {
```
coreylowman, line 106: Nice! We'll want to add some unit tests for this too

src/tensor_ops/matmul.rs
```diff
@@ -311,6 +411,47 @@ fn mm_atct<const M: usize, const K: usize, const N: usize>(
...
 
+/// batch matrix multiply `c += a * b`
+fn bmm<const M: usize, const K: usize, const N: usize, const B: usize>(
+    a: &[[[f32; K]; M]; B],
+    b: &[[f32; N]; K],
+    c: &mut [[[f32; N]; M]; B],
+) {
+    let b = b.as_ptr() as *const f32;
+
+    // Purely sequential for now, should parallelize using rayon or some other BLAS solution
```
coreylowman, line 422: Can open an issue for this as well to track

coreylowman (changes requested):
This is AWESOME! Thanks for contributing this!! Mostly just need to add tests for this new functionality - I can help out with that if you want.

Also, I'm curious if you had any thoughts for improving this more internal code. Did you find it easy to understand/work with?

src/tensor_ops/matmul.rs
jafioti: Couldn't either lol. Spent a while figuring out I just needed those. Probably some compiler bug.

coreylowman (approved):
Nice! 🎉

PR #106, opened 2022-07-20
Removing extra allocation in dropout

coreylowman:


coreylowman:
Actually this can be done with #105

PR #107, opened 2022-07-20
#67 `Optimizer::update` now returns `Result<(), UnusedParamsError>`

### LINKED ISSUE #67: If param is not used in loss calculation, `Optimizer::update()` causes runtime panic
Contrived example:

```rust
fn main() {
    let mut model: (Linear<5, 5>, Linear<5, 4>) = Default::default();
    let y = model.1.forward(Tensor1D::zeros().traced());
    let gradients = y.square().mean().backward();
    let mut opt: Sgd = Default::default();
    opt.update(&mut model, gradients);
}
```

coreylowman:
Closes #67


src/optim/adam.rs
```diff
@@ -107,15 +107,17 @@ impl<M> GradientProvider for Adam<M> {
...
 }
 
 impl<M: CanUpdateWithGradients> Optimizer<M> for Adam<M> {
     fn update(&mut self, module: &mut M, gradients: Gradients) {
         self.t = self.t.checked_add(1).unwrap();
         self.gradients = gradients;
-        module.update(self);
+        module
+            .update(self)
+            .expect("Parameter wasn't used for gradient computation");
```
coreylowman, line 120: This causes a runtime panic.

Idea: module.update() should continue with the gradient updating, and maybe the error can contain a list of unused parameters? Then Optimizer.update() can return that result and user can decide whether they want to panic or not

coreylowman:
This is now a breaking change and CanUpdateWithGradients is now quite a bit harder to read. BUT now users can decide if unused parameters are an issue, and if so get a list of all of them.

src/gradients.rs
```diff
@@ -258,7 +309,7 @@ pub trait GradientProvider {
 /// }
 /// ```
 pub trait CanUpdateWithGradients {
```
coreylowman, line 311: Add details to docstring for this about what the Error should be

src/nn/linear.rs
```diff
@@ -32,9 +32,11 @@ pub struct Linear<const I: usize, const O: usize> {
...
 
 impl<const I: usize, const O: usize> CanUpdateWithGradients for Linear<I, O> {
-    fn update<G: GradientProvider>(&mut self, grads: &mut G) {
-        self.weight.update(grads);
-        self.bias.update(grads);
+    fn update<G: GradientProvider>(&mut self, grads: &mut G) -> Result<(), UnusedParamsError> {
+        let mut unused = Ok(());
+        unused.union(self.weight.update(grads).map_err(|l| l.prepend("weight")));
+        unused.union(self.bias.update(grads).map_err(|l| l.prepend("bias")));
+        unused
```
coreylowman, line 39: This block feels like it could be improved somehow. the main core of it (self.weight.update & self.bias.update) is such a small portion of the code here.

vikigenius: Does this mean that any user written module will have to manually map the error if they want to include it?

Also is there no qualifying name here? What happens when there are multipe layers that have the same parameter name?

coreylowman: Yes as this PR is currently written, users would have to do the `map_err` thing and `union` if they wanted to properly compute the names of unused parameters.

How it works is the `l.prepend(...)` prepends the string to all parameters currently in `UnusedParamsError`. The union does a extend on the underlying vec. So in this case let's say both weight & bias were unused, it'd have 2 unused params. When this update call steps back up to the callee, it would also call .prepend() which would add more info to the parameters.

The example I was testing with was:
```rust
use dfdx::prelude::*;

fn main() {
    type Model = ((Repeated<Linear<5, 5>, 3>, ReLU), Linear<5, 4>);
    let mut model: Model = Default::default();

    // only use half of the model during forward
    let y = model.1.forward(Tensor1D::zeros().traced());
    let gradients = y.square().mean().backward();

    let mut opt: Sgd<Model> = Default::default();
    opt.update(&mut model, gradients).expect("");
}
```

Which fails with the message:
```
thread 'main' panicked at ': UnusedParamsError { param_locations: ["0.0.0.weight", "0.0.0.bias", "0.0.1.weight", "0.0.1.bias", "0.0.2.weight", "0.0.2.bias"] }', examples\bad.rs:14:39
error: process didn't exit successfully: `target\debug\examples\bad.exe` (exit code: 101)
```

Definitely isn't super easy to tell what parameters its talking about though...

vikigenius: Yes, but it is still better than having the same names repeated. This is one area where a language like Python with it's reflection shines, it can just lookup the name of the module class and prepend it.

But I say this looks good for now.

src/gradients.rs
```diff
@@ -101,6 +101,62 @@ impl Tape for NoneTape {
...
+
+pub trait CollectUnused<T> {
+    /// A method solely for adding two `Result<(), UnusedParamsError>` together.
+    ///
+    /// Cases:
+    /// 1. `self` and `b` are `Ok(())`; does nothing
+    /// 2. `self` is `Ok(())`, `b` is `Err(_)`; `*self = b`
+    /// 3. `self` is `Err(_)`, `b` is `Ok(())`; does nothing
+    /// 4. `self` and `b` are `Err(_)`; adds the unused params from `b` into `self`.
+    fn maybe_add_unused(&mut self, tag: &str, b: Self);
```
coreylowman, line 135: Still feel like this could be named better, not exactly sure what to call it. will brainstorm a bit

examples/custom.rs
```diff
@@ -22,10 +22,12 @@ impl<const IN: usize, const INNER: usize, const OUT: usize> ResetParams for Mlp<
 impl<const IN: usize, const INNER: usize, const OUT: usize> CanUpdateWithGradients
     for Mlp<IN, INNER, OUT>
 {
-    fn update<G: GradientProvider>(&mut self, grads: &mut G) {
-        self.l1.update(grads);
-        self.l2.update(grads);
-        self.relu.update(grads);
+    fn update<G: GradientProvider>(&mut self, grads: &mut G) -> Result<(), UnusedParamsError> {
+        let mut unused = Ok(());
+        unused.maybe_add_unused("l1.", self.l1.update(grads));
```
coreylowman, line 27: Rename unused to whatever the internals use

coreylowman:
TODO does this affect performance at all?

src/gradients.rs
```diff
@@ -248,18 +245,43 @@ pub trait GradientProvider {
...
-    fn update<G: GradientProvider>(&mut self, grads: &mut G);
+    /// Updates self given the [GradientProvider]. When any amount of parameters that
+    /// should be updated are NOT present in `G`, then this function should
+    /// populate the resulting [MissingGradients] struct with the names.
+    fn update<G: GradientProvider>(&mut self, grads: &mut G, missing: &mut MissingGradients);
+}
+
+/// An struct that holds names of parameters that were missing associated gradients.
+#[derive(Debug, Default)]
+pub struct MissingGradients {
```
coreylowman, line 257: rename to something like:
- `TensorsMissingGradients`
- `TensorGradientsMissing`
- `MissingTensorGradients`
- `MissingTensors`
- `UnusedParams`
- `ParamsMissingGradients`
- `NotUpdated`
- `TensorsNotUpdated`
- `ParamsNotUpdated`
- `NonUpdatedParams`

Issue #110, opened 2022-07-21
Dropout gradients


M1ngXU:
That [line](https://github.com/coreylowman/dfdx/blob/d7037549c3df0e0507bbe2269e7aee0f974324ba/src/tensor_ops/impl_dropout.rs#L26) sets the gradients to 0 if the `dropout(x)` at that index is greater than 0. Since the type is `f32`, x might already be negative or 0, but in that case the gradient shouldn’t be set to 0.

I am not 100% sure if I am right, though I’d probably create some tests with Tensors that are not `3.0`, but negative or zero.

coreylowman:
Great find, I think you're right!

M1ngXU:
Ok, will you change the code or should I do it later?
And also, how to find wheather a ‘value’ has been dropped out or x is just 0, as both would return 0?

coreylowman:
If you want to contribute the fix & test go for it! I don't think we can use map_df_uses_fx since we do need to know the difference between 0.0 and things that have been dropped out. You'll probably need to revert the change in this recent commit: https://github.com/coreylowman/dfdx/commit/053984e2a2358769fe9665c8d20475de5d223ddf. That may actually not have the same bug actually.

PR #113, opened 2022-07-21
Fix Issue #110 Fix (Dropout (test) for non-positive values)

M1ngXU:


M1ngXU:
Sorry, I somehow accidently commited a clippy warning fix into this. Should've created a new branch oops

src/lib.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(incomplete_features)]
```
coreylowman, line 1: Is this needed for something? After we remove this I can merge!

coreylowman (changes requested):

src/lib.rs
M1ngXU: Ok I removed that line (but now get a clippy warning)

coreylowman (approved):

src/lib.rs
coreylowman: Thanks! I think that's okay, should address that in a separate PR if want to fix

M1ngXU: what clippy config are you using? or are you getting this error too?

coreylowman: I get it when I do +nightly, since generic_const_exprs is incomplete, I'd rather keep that warning on. If you're getting the warning on stable then yeah I must be using different version.

```
> cargo clippy --version
clippy 0.1.62 (a8314ef7 2022-06-27)
```

M1ngXU: ok, i’ll check that later

Issue #96, opened 2022-07-20
Residual Connection with R(x) = F(x) + G(x) instead of R(x) = F(x) + x


M1ngXU:
In the ResNet paper, the residual connection of each downsampling block is not just `R(x) = F(x) + x`, but `R(x) = F(x) + G(x)` with G being a Conv2D with 1 filter.
This can probably be easily implemented by giving 2 modules for one Residual connection (and then just adding their values).

Edit: Working on it, though it might take longer because I first have to understand all the gradients/tape stuff xD

Edit2: I think I almost finished

coreylowman:
Closing this issue, thanks for the contribution @M1ngXU !

PR #118, opened 2022-07-23
Issue #96

M1ngXU:
See Issue #96


M1ngXU:
The `Module::forward` is quite inefficient with 3 allocations, so the old residual_add-connection is now residual_add. Also, I have no idea about the numpy stuff, so I am not sure how to fix the last numpy test (comments in code).

src/nn/residual.rs
```diff
@@ -1,52 +1,105 @@
...
+    T: Tensor<Dtype = f32, Tape = O::Tape>,
+    F: Module<T, Output = O>,
+    R: Module<T, Output = O>,
 {
-    type Output = F::Output;
+    type Output = O;
 
-    /// Calls forward on `F` and then adds `x` to the result: `F(x) + x`
+    /// Calls forward on `F` and `R` and then sums their result: `F(x) + R(x)`
     fn forward(&self, x: T) -> Self::Output {
```
coreylowman, line 49: I'm thinking this forward function can be simplified a lot with something like:
```rust
impl<F, R, T> Module<T> for Residual<F, R> {
where
    T: Tensor<Dtype = f32>,
    F: Module<T>,
    R: Module<T, Output = F::Output>,
{
    type Output = F::Output;
    fn forward(&self, x: T) -> Self:: Output {
        let (x, tape) = x.split_tape();

        // do R(x) on the tape
        let (r_x, tape) = self.1.forward(x.duplicate().put_tape(tape)).split_tape();

        // do F(x) on the tape
        let f_x = self.0.forward(x.put_tape(tape));

        add(f_x, &r_x)
    }
}
```

src/nn/residual.rs
```diff
@@ -1,52 +1,105 @@
...
-/// let module: Residual<ReLU> = Default::default();
+/// let module: Residual<ReLU, Square> = Default::default();
 /// let x = Tensor1D::new([-2.0, -1.0, 0.0, 1.0, 2.0]);
 /// let y = module.forward(x);
-/// assert_eq!(y.data(), &[-2.0, -1.0, 0.0, 2.0, 4.0]);
+/// assert_eq!(y.data(), &[4.0, 1.0, 0.0, 2.0, 6.0]);
 /// ```
 #[derive(Debug, Clone, Default)]
-pub struct Residual<F>(F);
+pub struct Residual<F, R>(F, R);
```
coreylowman, line 19: I like the idea of having both Residual & ResidualAdd. I imagine most people will probably use ResidualAdd, so let's rename them
- `Residual` does `f(x) + x`
- `GeneralizedResidual` does `f(x) + r(x)`

coreylowman (changes requested):
Thanks for the contribution!

src/nn/residual.rs
M1ngXU: What about `SkipConnection` instead of `GeneralizedResidual`?

src/nn/residual.rs
M1ngXU: ok, makes sense. i think that i am missing lots of differentiation basics :/

M1ngXU:
hmmm, file deleted and new file oof

src/nn/residual.rs
M1ngXU: ok, after re-reading some wiki pages i think that i understand it better now - but shouldn't the gradients of `F` and `R` have to be summed AFTER the backprop of `F` and `R` (so the addition would have to be pushed to the tape first?) or am I not understanding `binary_map` ?

src/nn/residual.rs
coreylowman: No worries. So far the modules in nn don't actually specify any backward ops, they just either call other module's forwards or call operations from tensor_ops.

src/nn/residual.rs
coreylowman: All the operations on the tape are executed in reverse, so for this function the add/binary_map backward op will be executed, then the f(x) backward op is executed using the gradient from add(), and then the r(x) backward ops using gradient from add(). Does that help?

Let's keep both Residual and SkipConnection, and then GeneralizedResidual. SkipConnection can just be a type alias (`type SkipConnection<F, R> = Residual<F, R>;`

M1ngXU: do you have an idea why the numpy test fails?

M1ngXU: > All the operations on the tape are executed in reverse, so for this function the add/binary_map backward op will be executed, then the f(x) backward op is executed using the gradient from add(), and then the r(x) backward ops using gradient from add(). Does that help?

but isn't the gradient to be used for f(x) and r(x) the same? and shouldn't the layer l before f and r get df + dr as the gradient?

M1ngXU: > Let's keep both Residual and SkipConnection, and then GeneralizedResidual. SkipConnection can just be a type alias (`type SkipConnection<F, R> = Residual<F, R>;`

Are you sure about creating a type alias? Then it would probably make sense to add an alias `Dense` for `Linear` ...

src/nn/generalized_residual.rs
```diff
@@ -0,0 +1,379 @@
...
+        &self,
+        filename_prefix: &str,
+        w: &mut zip::ZipWriter<W>,
+    ) -> zip::result::ZipResult<()>
+    where
+        W: std::io::Write + std::io::Seek,
+    {
+        // I have no idea what to put here
+        self.0.write(filename_prefix, w)?;
+        self.1.write(filename_prefix, w)?;
```
coreylowman, line 74: I think these are why the save/load is failing :) basically the first parameter is the prefix of the file to write the data to. Since they both are using `filename_prefix`, they are overwriting each other!

You can use something like this for the first param of both lines: `&format!("{filename_prefix}0.")` and `&format!("{filename_prefix}1.")`, which will save the first param to `0.<submodule>.npy` and `1.<submodule>.npy`.

Same comment for LoadFromNpz

M1ngXU: thanks for the help! it works now

M1ngXU:
should be ready for merge now

M1ngXU:
finally the tests all passed 😅

M1ngXU:
with the new `GeneralizedResidual`, the `Residual` struct can be removed, if we add a `Linear`-Activation, so f(x) = x, and then `type Residual<F> = GeneralizedResidual<F, Linear>`

src/nn/residual.rs
M1ngXU: > Does that help?

but what should be added for the gradients? and what about the layer before f and r? shouldn't the gradient there be the sum of df and dr?

coreylowman: The gradients are handled automatically by calling add (assuming F & R properly add their derivatives & backward ops to the tape). That's the nice thing about all the stuff in nn - none of it needs to know anything about gradients. All of the functions in tensor_ops handle gradient related things, nn just pieces them together.

This is the key piece about auto diff - you can implement gradient stuff for the lower level building blocks (like add, mul, apply some function, etc), and then piece them together automatically without having to explicitly handle the gradients.

Does that help?

coreylowman (approved):

coreylowman:
@M1ngXU this looks good to me for merging - can you update or rebase your branch (e.g. the dropout change was already merged into main)

src/nn/residual.rs
M1ngXU: i know that, but wouldn't the layer in the forward pass before the residual connection require the addition of the gradients of the residual connection? so the gradient for the layer before the residual connection needs a backward_op pushed _before_ the backward_ops are pushed for the residual layer? how is this done in add?

M1ngXU:
> @M1ngXU this looks good to me for merging - can you update or rebase your branch to current master (e.g. the dropout change was already merged into main)

how exactly do i do that? 😅

src/nn/residual.rs
coreylowman: Yeah so the layer before has a gradient of `derivative * result_grad`, where result_grad is calculated by this residual layer during the backward pass (and is the gradient of the input tensor in the forward method). That's basically handled by each tensor having an id associated with it, and the id is used to retrieve the gradient during the backward pass. So the result tensor from the layer before has the same id as the input into the residual layer.

coreylowman:
> > @M1ngXU this looks good to me for merging - can you update or rebase your branch to current master (e.g. the dropout change was already merged into main)
> 
> how exactly do i do that? 😅

This article should walk you through it https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/syncing-a-fork

src/nn/residual.rs
M1ngXU: > where result_grad is calculated by this residual layer during the backward pass (and is the gradient of the input tensor in the forward method).

so main_forward + residual_forward?
in which line is this done 😅 ?

M1ngXU:
> > > @M1ngXU this looks good to me for merging - can you update or rebase your branch to current master (e.g. the dropout change was already merged into main)
> > 
> > 
> > how exactly do i do that? 😅
> 
> This article should walk you through it https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/syncing-a-fork

ok, should work now. i thought that i already tried that, but i appearently synced main and not BetterResidual.

src/nn/residual.rs
coreylowman: Oh wait I may have misinterpreted, when you say layer before residual connection, are you talking about F.forward(x)/R.forward(x), or the module before Residual<F, R>?

M1ngXU:
> with the new `GeneralizedResidual`, the `Residual` struct can be removed, if we add a `Linear`-Activation, so f(x) = x, and then `type Residual<F> = GeneralizedResidual<F, Linear>`

@coreylowman how about this?

coreylowman:
> with the new `GeneralizedResidual`, the `Residual` struct can be removed, if we add a `Linear`-Activation, so f(x) = x, and then `type Residual<F> = GeneralizedResidual<F, Linear>`

I like the idea about moving Residual to use GeneralizedResidual, but I think we'd need something like `Identity` (which just returns x in the forward pass) instead of Linear for R.

M1ngXU:
> > with the new `GeneralizedResidual`, the `Residual` struct can be removed, if we add a `Linear`-Activation, so f(x) = x, and then `type Residual<F> = GeneralizedResidual<F, Linear>`
> 
> I like the idea about moving Residual to use GeneralizedResidual, but I think we'd need something like `Identity` (which just returns x in the forward pass) instead of Linear for R.

the name Identity makes more sense

src/nn/residual.rs
M1ngXU: > Oh wait I may have misinterpreted, when you say layer before residual connection, are you talking about F.forward(x)/R.forward(x), or the module before Residual<F, R>?

the module before Residual<F,R> in the forward pass (so the output of that module is the input of Residual<_, _>)

M1ngXU: > add_grad + f_grad + r_grad

but if the `add(y, &x)` is executed before `2.a` in the backprop, how can f_grad be fetched from gradients to calculate `x`'s gradients for `1`?

coreylowman: The three are summed together independently. After add's backprop x's gradient is only 1/3 complete (add_grad). Once f's backward op is executed then x is 2/3 complete (add_grad + f_grad), then finally after r's backward op x is complete (add_grad + f_grad + r_grad).

It works because x is input for add, f, and r, not the result. Gradients only need to be fully "complete" for backward ops in which they are the result.

coreylowman: it depends on what f & r are. for binary map the call to addmul does `rhs_grad += rhs_deriv * result_grad`, so if `rhs_grad` is x's grad that's where part of the accumulation happens.

If F and R were both using `map` for instance then both accumulations would happen in [map](https://github.com/coreylowman/dfdx/blob/main/src/tensor_ops/map.rs#L235) (where x's gradient is `g`)

M1ngXU: > Linear<10, 5>

i mean that linear layer, what gradients will it use? it will be `lhs_grad`, won't it? but when is `rhs_grad` added to `lhs_grad`?

PR #120, opened 2022-07-25
Add Transformers!!!

jafioti:
This PR adds MultiHeadAttention, transformer blocks, and a transformer encoder. It also contains a new reshaping trait because the old one wasn't working when implementing attention. These two reshape techniques should be merged. 

This PR also simplifies the batch matmul by using mm underneath in a loop rather than define custom functions for bmm.

Additional work will have to be done to add a transformer decoder and a full seq2seq transformer, but should be fairly straightforward given this base.


src/nn/split_into.rs
```diff
@@ -17,7 +17,7 @@ use crate::prelude::*;
 /// let _: (Tensor1D<3>, Tensor1D<7>) = model.forward(Tensor1D::<5>::zeros());
 /// ```
 #[derive(Debug, Default, Clone)]
-pub struct SplitInto<T>(T);
+pub struct SplitInto<T>(pub(crate) T);
```
coreylowman, line 20: we can probably just make this `pub`. I think all the nn modules can just pub all their underlying modules.

src/nn/transformer.rs
```diff
@@ -0,0 +1,1097 @@
...
+    const M: usize,
+    const N: usize,
+    const K: usize,
+    const V: usize,
+    const H: usize,
+> {
+    w_q: Linear<N, K>,
+    w_k: Linear<M, K>,
+    w_v: Linear<M, V>,
+    w_o: Linear<V, M>,
```
coreylowman, line 27: I think these all need to be multiplied by number of heads (`H`), right? Otherwise there's only just 1 head?

src/nn/transformer.rs
```diff
@@ -0,0 +1,1097 @@
...
+        self.w_q.update(grads);
+        self.w_k.update(grads);
+        self.w_v.update(grads);
+        self.w_o.update(grads);
+    }
+}
+
+/// Normal self attention (where same tensors are used for keys, queries and values)
+impl<const M: usize, const K: usize, const V: usize, const S: usize, const H: usize>
+    Module<Tensor2D<S, M>> for MultiHeadAttention<M, M, K, V, H>
```
coreylowman, line 54: We'll need to add a `TAPE: Tape` generic arg here and add to Input/Output. Also will need to do some tape manipulation in the actual forward functions to make sure all the ops are on the tape

src/nn/transformer.rs
```diff
@@ -0,0 +1,1097 @@
...
+}
+
+/// A single transformer block containing self attention, feed forward and layer norms
+#[derive(Debug, Clone, Default)]
+pub struct TransformerBlock<const M: usize, const I: usize, const K: usize, const H: usize> {
+    attn: MultiHeadAttention<M, M, K, M, H>,
+    norm1: LayerNorm1D<M>,
+    norm2: LayerNorm1D<M>,
+    ff: (Linear<M, I>, ReLU, Linear<I, M>),
+}
```
coreylowman, line 210: Could this be implemented as
```rust
type TransformerBlock<....> = (
    Residual<MultiHeadAttention<...>>,
    LayerNorm1D<M>,
    Residual<(Linear, ReLU, Linear)>,
)
```

I was envisioning using `Residual` a bunch at this level.

src/nn/transformer.rs
```diff
@@ -0,0 +1,1097 @@
...
+/// - `L` The number of layers.
+/// - `H` The number of heads for self attention.
+/// TODO: Doctests
+#[derive(Debug, Clone)]
+pub struct TransformerEncoder<const M: usize, const I: usize, const L: usize, const H: usize>
+where
+    Assert<{ M % H == 0 }>: ConstTrue,
+{
+    blocks: Repeated<TransformerBlock<M, I, M, H>, L>,
+}
```
coreylowman, line 290: Similarly I think this could just be
```rust
type TransformerEncoder<...> = Repeated<TransformerBlock<...>, L>;
```

src/nn/transformer.rs
```diff
@@ -0,0 +1,1097 @@
...
+        let mut rng = thread_rng();
+        let inp: Tensor1D<4> = Tensor1D::rand(&mut rng);
+        let out: Tensor2D<2, 2> = inp.duplicate().reshape();
+        println!("Inp: {:?}", inp);
+        println!("Out: {:?}", out);
+        //panic!("")
+    }
+
+    #[test]
+    fn test_transformer_encoder() {
```
coreylowman, line 544: I've been meaning to do this for a while, but i'd like to take all the actual floats out of the tests so they are easier to scan. Don't need to address in this PR, but probably shortly after. I think we can take advantage of the numpy::load functions here and store .npy files somewhere in the repo.

src/tensor_ops/mod.rs
```diff
@@ -79,3 +79,8 @@ mod reshape;
 
 #[cfg(feature = "nightly")]
 pub use reshape::*;
+
+#[cfg(feature = "nightly")]
+mod impl_reshape;
+#[cfg(feature = "nightly")]
+pub use impl_reshape::*;
```
coreylowman, line 86: Want to just replace the contents of the reshape file? You weren't able to use the reshape function in there right?

coreylowman (changes requested):
Awesome contribution, this is an amazing start

src/nn/transformer.rs
jafioti: No, there are H heads but each is M / H dims, so all heads combined are M dims. The projection is combined for all heads, and the head splitting only happens after. Take a look at the minGPT implementation: https://github.com/karpathy/minGPT/blob/master/mingpt/model.py

src/nn/transformer.rs
jafioti: When passing tensors to matmul, why does the RHS have a NoneTape? Don't the gradients flow to both operators?

src/nn/transformer.rs
jafioti: I would want to do that, but notice the where statement. Technically this could be gone, but I think it's nice that it checks that the embedding dims are dividable by the num of heads when the model is created, rather than when forward is called.

src/nn/transformer.rs
jafioti: Awesome!

src/nn/transformer.rs
jafioti: I like that idea. I'm going to leave it for a seperate PR though.

src/tensor_ops/mod.rs
jafioti: I'll get rid of the reshape file. I feel like impl_reshape stays consistent with all the other impl files since it's a trait now, rather than a generic function.

src/nn/transformer.rs
M1ngXU: > When passing tensors to matmul, why does the RHS have a NoneTape? Don't the gradients flow to both operators?

i think because the gradients of rhs are stored in the tape of lhs

jafioti:
At this point I think I've added everything I wanted for this PR. Let me know if anything else needs to change, otherwise its ready to merge.

src/nn/transformer.rs
coreylowman: Exactly yep, just a convention that I went with - rhs will never own the tape in all the binary ops, lhs will either have an OwnedTape or NoneTape. the tape is extracted from lhs and used to add ops to for both operators.

For tracing, none of the stored parameters in a module will ever actually own the tape, the tape is always inserted into the input into the network and moved to intermediate results. Just so happens that all the nn operations so far have parameters as rhs arguments.

src/nn/transformer.rs
coreylowman: I see what you're saying, but I don't think that alone is worth all the extra code & tests that come with it.

src/nn/transformer/decoder.rs
```diff
@@ -0,0 +1,233 @@
...
+    Assert<{ H * S1 * (M / H) == S1 * M }>: ConstTrue,
+{
+    type Output = (Tensor2D<S1, M, T>, T);
+
+    fn forward(&self, (input, from_enc): (Tensor2D<S1, M, T>, Tensor2D<S2, N, T>)) -> Self::Output {
+        let (input, input_tape) = input.split_tape();
+        let (x, enc_tape) = self
+            .attn
+            .forward((input.duplicate().put_tape(input_tape), from_enc));
+        (self.ff.forward(add(x, &input)), enc_tape)
```
coreylowman, line 81: I think this should be `self.ff.forward(add(x.put_tape(enc_tape), &input)`, and then can remove the tape from the output tuple?

src/nn/transformer/base.rs
```diff
@@ -0,0 +1,221 @@
...
+    > Module<(Tensor2D<S1, M, T>, Tensor2D<S2, N, T>)> for MultiHeadAttention<M, N, K, V, H>
+where
+    Assert<{ S1 * K == H * S1 * (K / H) }>: ConstTrue,
+    Assert<{ S2 * K == H * S2 * (K / H) }>: ConstTrue,
+    Assert<{ S2 * V == H * S2 * (V / H) }>: ConstTrue,
+    Assert<{ H * S1 * (V / H) == S1 * V }>: ConstTrue,
+{
+    type Output = (Tensor2D<S1, M, T>, T);
+
+    fn forward(&self, (input, from_enc): (Tensor2D<S1, M, T>, Tensor2D<S2, N, T>)) -> Self::Output {
```
coreylowman, line 108: So generally for tape, there should only be 1 on the input & one on the output. And at least 1 tensor in every operation in forward needs to own `T`. so all the reshape ops should have tape, and then the tape needs to be in  `queries` when input to the matmul argument.

Then the output of this can just be `type Output = Tensor2D<S1, M, T>`, as that output tensor owns `T`, so its not separate

src/nn/transformer/decoder.rs
```diff
@@ -0,0 +1,233 @@
...
+/// A transformer decoder block. Different than the normal transformer block as this self attention accepts an additional sequence from the encoder.
+///
+/// # Generics
+/// - `M` The embedding size of token vectors from decoder.
+/// - `N` The embedding size of token vectors from encoder.
+/// - `K` The size of the keys in self attention.
+/// - `H` The number of attention heads.
+/// TODO: Doctests
+#[derive(Default, Debug)]
+pub struct TransformerDecoderBlock<
```
coreylowman, line 14: This one can't be just written as a sequential module right? Because the residual would receive a tuple argument, and just needs to keep the .0 of that tuple to add to the MHA output later?

jafioti: Correct, the from_encoder tensor only gets sent to the attention, not added at the end.

jafioti:
@coreylowman Can this get merged? Is there anything else in the way?

src/nn/transformer/base.rs
```diff
@@ -0,0 +1,221 @@
...
+/// # Generics
+/// - `M` The embedding size of token vectors from decoder.
+/// - `N` The embedding size of token vectors from encoder.
+/// - `K` The size of the keys in self attention.
+/// - `V` The size of the values.
+/// - `H` The number of attention heads.
+///
+/// # Examples
+/// `MultiHeadAttention<8, 10, 10, 10, 2>` is an attention layer with 2 heads and 10 token, key and value dims.
+/// TODO: Doctests fail for some reason
```
coreylowman, line 15: I was seeing this too on nightly, unsure why it happened and was unable to reproduce in regular tests

coreylowman (approved):
Looks great from my end! Thanks again for all the work

PR #124, opened 2022-07-29
Add Conv2d

### LINKED ISSUE #1: Convolution mega issue

coreylowman:
Closes #1 

- [x] Add tests for conv2d & batch_conv2d
- [x] Add tests for Conv2D layer
- [x] Add example file
- [x] Clean up implementation


src/tensor_ops/conv.rs
```diff
@@ -0,0 +1,302 @@
+use crate::prelude::*;
+
+pub fn conv2d<
```
coreylowman, line 3: Adding a `const CFG: ConvConfig` could be really helpful to reduce const generic parameters here. Will require an additional feature to enable usage of structs in const

src/tensor_ops/conv.rs
```diff
@@ -0,0 +1,302 @@
...
+    const H: usize,
+    const W: usize,
+    const OH: usize,
+    const OW: usize,
+>(
+    img: &[[[f32; W]; H]; C],
+    weight: &[[[[f32; K]; K]; C]; OC],
+    bias: &[f32; OC],
+    out: &mut [[[f32; OW]; OH]; OC],
+) {
```
coreylowman, line 175: For anyone who looks at this PR and wonders about this implementation, this was actually faster in my local tests than im2col & matmul for a variety of batch/image sizes.

src/tensor_ops/conv.rs
M1ngXU: this could be really helpful, especially if the amount of filters can be set independently, so there could be a default convconfig for a kernel size of 3x3 with 0 padding etc. And conv2d is nightly anyway, so another feature wouldn't hurt much

M1ngXU: a default option like this could use typealiasing though and would not require extra-features

src/nn/conv.rs
```diff
@@ -0,0 +1,256 @@
...
+    #[test]
+    fn test_3_conv_sizes() {
+        type A = Conv2D<1, 2, 3>;
+        type B = Conv2D<2, 4, 3>;
+        type C = Conv2D<4, 1, 1, 1, 1>;
+
+        type Img = Tensor3D<1, 10, 10>;
+        let _: Tensor3D<2, 8, 8> = A::default().forward(Img::zeros());
+        let _: Tensor3D<4, 6, 6> = <(A, B)>::default().forward(Img::zeros());
+        let _: Tensor3D<1, 8, 8> = C::default().forward(Tensor3D::<4, 6, 6>::zeros());
```
coreylowman, line 212: So something with generic const exprs is currently broken. If you remove this line you will get a bunch of linker errors about symbols not being included. This makes convs pretty unusable at this point 😞 .

vikigenius: Yup, I could reproduce this as well. But even apart from the extern symbols there is also something weird going on with the module.

```rust
  = note: /usr/bin/ld: /home/void/Projects/Rust/dfdx/target/debug/deps/dfdx-5aab129f28193da6.5gs3y0we70clq0q0.rcgu.o: in function `dfdx::nn::impl_module_for_tuples::<impl dfdx::nn::module::Module<Input> for (A,B,C)>::forward':
          /home/void/Projects/Rust/dfdx/src/nn/impl_module_for_tuples.rs:79: undefined reference to `<dfdx::nn::conv::Conv2D<_,_,_,_,_> as dfdx::nn::module::Module<dfdx::tensor::structs::Tensor3D<_,_,_,TAPE>>>::forward'
          /usr/bin/ld: /home/void/Projects/Rust/dfdx/target/debug/deps/dfdx-5aab129f28193da6: hidden symbol `_ZN155_$LT$dfdx..nn..conv..Conv2D$LT$_$C$_$C$_$C$_$C$_$GT$$u20$as$u20$dfdx..nn..module..Module$LT$dfdx..tensor..structs..Tensor3D$LT$_$C$_$C$_$C$TAPE$GT$$GT$$GT$7forward17h4239bcee10941f68E' isn't defined
          /usr/bin/ld: final link failed: bad value
```

coreylowman: This is fixed on the latest nightly

Issue #121, opened 2022-07-26
Embedding Layers


jafioti:
In NLP, tokens are converted from tensors of `usize` to tensors of `f32`, where each `usize` is an index into a "library" of token embedding vectors. I'm pretty sure the inner workings of this work by converting each `usize` into a one-hot vector (for instance, 2 can be converted to `[0, 0, 1, 0, 0, ...]`) and then multiplied by the embedding weight matrix. Since only one value in the vector is 1, only one of the vectors in the weight matrix will be "selected".

coreylowman:
I was thinking this could be an Tensor2D<M, N>, where indices can be between 0 and M, and then the `usize` is just used to index into that tensor to grab the m'th row (resulting in a 1D<N>). could add a `gather` operation for this I think

jafioti:
How would the gather operation work? I was thinking turning indexes into one hots and doing matmuls is too inefficient, all but one of the embeds would be multiplied by 0, so I was going to just sequentially grab the vectors. Not sure how to move them to the output tensor though and keep the tape intact.

coreylowman:
It'd be similar to the `gather_last_dim()` function but I guess instead it would be gathering from the first dimension?

```rust
let embeddings: Tensor2D<10, 3> = Tensor2D::zeros();
let batch_embeddings: Tensor2D<2, 3> = gather(embeddings, &[4, 6], tape);
```
?

coreylowman:
This should now be possible via the `.select()` method of `Select1` trait. Should be something close to:

```rust
struct Embeddings<const N: usize, const D: usize> {
    data: Tensor2D<N, D>,
}

impl<const N: usize, const D: usize, const Z: usize> Module<[usize; Z]> for Embeddings<N, D> {
    type Output = Tensor2D<Z, D, OwnedTape>;
    fn forward(&self, input: [usize; Z]) -> Self::Output {
        self.data.trace().select(&input)
    }
}
```

Will need to figure out where the tape comes from (e.g. the above impl creates a tape inside, but not sure that's what it should be). May want to pass tape as input (maybe tuple of indices & tuple)

coreylowman:
There should also be a `PaddedEmbedding` which is equivalent to using the pad_idx in pytorch

coreylowman:
Posting partial progress here. One thing I need to figure out is the padding index, which has 0 embeddings and always has gradient of 0. It's easy to initialize it to 0, but unsure of how to mask out specific indices. Perhaps some masking operation?

[...]

PR #138, opened 2022-08-12
Broadcasting, reductions, and select rework

### LINKED ISSUE #114: Mean, Sum across arbitrary dimensions (especially batch dimension)
This might potentially depend on Reshape, but it would be good to compute mean/sum across arbitrary dimensions by letting mean/sum take `dim` or `axis` as an argument just like how Pytorch does it.

This should also be useful for the `batch_norm` issue https://github.com/coreylowman/dfdx/issues/78 because currently there is no api/easy way to compute the mean across the batch dimension.

### LINKED ISSUE #137: Add broadcasting tensor op and remove the binary_op broadcasting methods

### LINKED ISSUE #139: Select operation
Should support the following:
- `Tensor3D::<3, 5, 7>::select_axis::<0>([usize; 1])` => `Tensor3D<1, 5, 7>`
- `Tensor3D::<3, 5, 7>::select_axis::<0>(usize)` => `Tensor2D<5, 7>`
- `Tensor3D::<3, 5, 7>::select_axis::<-1>([[usize; 5]; 3])` => `Tensor2D<3, 5>`
- `Tensor3D::<3, 5, 7>::select_axis::<-1>([[[usize; 2]; 5]; 3])` => `Tensor2D<3, 5, 2>`
- etc

coreylowman:
Closes #137, Closes #114, Closes #139 

Enables #121 

This is a giant PR with many features & changes, so i'll probably leave this up for a bit while I review. This is in a working state though and at this point just needs some documentation work.

Summary of changes:
- Removed all `*_last_dim()` functions.
- Add `*_axis()` functions that take a const isize that indicate the dimension to reduce.
- Add `Broadcast[1-4]` traits, which include methods to broadcast up to 4 dimensions larger
- Add `Select1` tensor trait, that replaces gather_last_dim and is much more general
- Remove broadcast binary map methods
- Add Reduce1Axis, ForEachBroadcast[1-4] traits for device
- Add `HasAxis<I>` trait for arrays that enables getting size of axis.


src/tensor_ops/broadcast.rs
```diff
@@ -0,0 +1,265 @@
+//! TODO describe implementation details of why there are 3 versions of the traits and how they differ.
```
coreylowman, line 1: It's pretty confusing to see multiple versions of broadcast functions in ide completion. Ideally people should NOT be using ConstBroadcast or BroadcastTo, however those exist to make generic functions possible.

BroadcastTo exists for some of *_axis methods that need to immediately broadcast reduced values back to the original size (std_axis, normalize, log_softmax).

ConstBroadcast exists to specify the associated type that is used to implement Broadcast[1-4].

src/tensor_ops/broadcast.rs
```diff
@@ -0,0 +1,265 @@
...
+
+use super::utils::move_tape_and_add_backward_op;
+use crate::prelude::*;
+
+pub trait ConstBroadcast1<const I: isize, const M: usize> {
```
coreylowman, line 6: Maybe rename to KnownBroadcast?

src/tensor_ops/broadcast.rs
coreylowman: BroadcastTo could potentially be removed if variants of the _axis functions exist that don't actually reduce or change dimension. e.g. `sum_axis_rebroadcast<T, const I: isize>(t: T) -> T`

coreylowman: Or another option is to remove Broadcast[1-4] in favor of using BroadcastTo everywhere. I think this option seems feasible

coreylowman: Removed everything in favor of juts BroadcastTo, which is the "opposite" of the current Reduce1

src/tensor_ops/reduce.rs
```diff
@@ -0,0 +1,48 @@
...
+// 2d
+reduction!(Tensor2D, [M, N], Tensor1D, [N], 0);
+// reduction!(Tensor2D, [M, N], Tensor1D, [M], 1);
+reduction!(Tensor2D, [M, N], Tensor1D, [M], -1);
+
+// 3d
+reduction!(Tensor3D, [M, N, O], Tensor2D, [N, O], 0);
+reduction!(Tensor3D, [M, N, O], Tensor2D, [M, O], 1);
+// reduction!(Tensor3D, [M, N, O], Tensor2D, [M, N], 2);
+reduction!(Tensor3D, [M, N, O], Tensor2D, [M, N], -1);
```
coreylowman, line 41: Having multiple impls for the same axis (i.e. the last axis can be 2 or -1), means the compiler complains about multiple trait impls that require explicitly setting that.

I.e. with both -1 and 2 available for reducing/broadcasting:
```rust
let a: Tensor2D<3, 5> = ...;
let b: Tensor3D<3, 5, 7> = a.broadcast1(); // compiler will complain about multiple impls here
let c= Broadcast1<Tensor3D<3, 5, 7>, -1>::broadcast1(b); // this will be fine
```

However if just -1 is available (or 2) then you can use the `.broadcast1()`.

src/tensor_ops/mod.rs
```diff
@@ -10,69 +10,85 @@
...
 //!
 //! See relevant functions for more examples.
 //!
 //! # Broadcasts
 //!
-//! Some binary functions need to broadcast one argument to be the same size as the other (e.g. [add_broadcast_rhs_last()]).
-//! These methods are named as `<operation>_broadcast_<argument>_<dimension>()`. Currently all of the functions
-//! broadcast the second argument (`rhs`). And there are version where the first dimension is broadcast and the last dimension
-//! is broadcast
+//! Broadcasting tensors is provided through the following traits:
```
coreylowman, line 24: Provide a complete list of possible broadcasts?

src/tensor_ops/mod.rs
```diff
@@ -10,69 +10,85 @@
...
 //! # Reductions
 //!
-//! There are a number of functions that reduce a dimension (e.g. [mean_last_dim()]).
-//! These functions are all labeled with `*_last_dim()` at the end.
+//! There are a number of functions that reduce a dimension (e.g. [mean_axis()]).
+//! These functions are all labeled with `*_axis()` at the end.
 //!
 //! Reducing a dimension means removing that dimension from the tensor by reducing it to 1 number.
-//! For example calling [sum_last_dim()] on a `Tensor2D<2, 5>` would result in a `Tensor1D<2>`.
+//! For example calling [sum_axis::<1>()] on a `Tensor2D<2, 5>` would result in a `Tensor1D<2>`.
```
coreylowman, line 17: Change to `sum_axis::<-1>()`

src/tensor_ops/reduce.rs
```diff
@@ -0,0 +1,48 @@
+use crate::prelude::*;
+
+/// TODO
+pub trait Reduce1<const I: isize>: Tensor<Dtype = f32> {
+    type Reduced: Tensor<Dtype = Self::Dtype, Tape = Self::Tape> + Broadcast1<Self, I>;
+    type DeviceR: Reduce1Axis<Self::Array, <Self::Reduced as HasArrayType>::Array, I>;
```
coreylowman, line 6: Having a separate device here is necessary because the normal Tensor::Device doesn't have a bound for Reduce1Axis<_, _, I> (and this isn't possible since it doesn't have axis to `I`).

If this new device wasn't specified then any generic function that uses Reduce1 would have to also have T::Device: Reduce1Axis<_, _, I>, which I felt was pretty verbose.

src/tensor_ops/impl_normalize.rs
```diff
@@ -1,25 +1,28 @@
...
 /// # use dfdx::prelude::*;
 /// let a = Tensor1D::new([-2.0, -1.0, 0.0, 5.0, 2.0]);
 /// let r = normalize(a, 1e-5); // or a.normalize(1e-5);
-/// assert!(mean_last_dim(r.duplicate()).data().abs() < 1e-6);
-/// assert!((std_last_dim(r.duplicate(), 0.0).data() - 1.0).abs() < 1e-6);
+/// assert!(r.clone().mean_axis::<-1>().data().abs() < 1e-6);
+/// assert!((r.clone().std_axis::<-1>(0.0).data() - 1.0).abs() < 1e-6);
 /// ```
-pub fn normalize<T: Tensor<Dtype = f32>>(t: T, epsilon: T::Dtype) -> T {
+pub fn normalize<T: Reduce1<-1>>(t: T, epsilon: T::Dtype) -> T {
```
coreylowman, line 16: Should this be normalize_axis?

src/devices/reduce_axis.rs
```diff
@@ -0,0 +1,238 @@
...
+                for p in 0..P {
+                    out[n][o][p] = inp.iter().map(|x| x[n][o][p]).reduce(f).unwrap();
+                }
+            }
+        }
+    }
+}
+
+macro_rules! reduce1_nonzero_axis {
+    ($Idx:expr, $SubIdx:expr, $Reduced:ty, $SubArrTy:ty, {$($Vs:tt),*}) => {
```
coreylowman, line 110: Change SubArrTy to ArrTy

src/devices/select.rs
```diff
@@ -0,0 +1,104 @@
...
+//! ## Selecting either 1 or Z elements from a non-zero axis
+//!
+//! Then all three arrays with have the same dimension as the 0th axis.
+//! Do a for loop over the 0th axis and recurse!
+
+use super::{Cpu, ForEachElement};
+use crate::arrays::CountElements;
+
+/// Select values from `T` using `Indices` and producing `R` along a single `AXIS`.
+pub trait SelectAlongAxis<T: CountElements, Indices, R: CountElements, const AXIS: isize> {
```
coreylowman, line 23: Need to add tests for this

src/tensor_ops/impl_normalize.rs
coreylowman: Probably, but not in this PR

PR #153, opened 2022-08-21
Added draft BatchNorm1D implementation

vikigenius:
This is a WIP draft PR for batch normalization, wanted to discuss some potential design choices and issues.

### DIFF
--- a/src/nn/batch_norm.rs
+++ b/src/nn/batch_norm.rs
@@ -1,0 +1,193 @@
+use crate::prelude::*;
+
+
+pub struct BatchNorm1D<const C: usize> {
+    pub gamma: Tensor1D<C, NoneTape>,
+    pub beta: Tensor1D<C, NoneTape>,
+    pub momentum: f32,
+    pub epsilon: f32,
+    pub running_mean: Tensor1D<C, NoneTape>,
+    pub running_var: Tensor1D<C, NoneTape>,
+}
+
+
+impl<const C: usize> Default for BatchNorm1D<C> {
+    /// Fills [Self::gamma] with 1s and [Self::beta] with 0s, sets [Self::momentum] to `0.1`,
+    /// sets [Self::epsilon] to `1e-5`
+    fn default() -> Self {
+        Self {
+            gamma: Tensor1D::ones(),
+            beta: Tensor1D::zeros(),
+            momentum: 0.1,
+            epsilon: 1e-5,
+            running_mean: Tensor1D::zeros(),
+            running_var: Tensor1D::ones(),
+        }
+    }
+}
+
+
+impl<const C: usize> ResetParams for BatchNorm1D<C> {
+    /// Fills [Self::gamma] with 1s and [Self::beta] with 0s.
+    fn reset_params<R: rand::Rng>(&mut self, _: &mut R) {
+        Cpu::fill(self.gamma.mut_data(), &mut |v| *v = 1.0);
+        Cpu::fill(self.beta.mut_data(), &mut |v| *v = 0.0);
+        Cpu::fill(self.running_mean.mut_data(), &mut |v| *v = 0.0);
+        Cpu::fill(self.running_var.mut_data(), &mut |v| *v = 1.0);
+    }
+}
+
+
+impl<const C: usize> CanUpdateWithGradients for BatchNorm1D<C> {
+    /// Updates [Self::gamma] and [Self::beta].
+    fn update<G: GradientProvider>(&mut self, grads: &mut G, unused: &mut UnusedTensors) {
+        self.gamma.update(grads, unused);
+        self.beta.update(grads, unused);
+    }
+}
+
+
+impl<const C: usize> Module<Tensor1D<C, NoneTape>> for BatchNorm1D<C> {
+    type Output = Tensor1D<C, NoneTape>;
+
+    /// Forward implementation for NoneTape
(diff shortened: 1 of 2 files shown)


vikigenius:
Since the behavior of batchnorm is different for training and inference we need some way to test for both.

One possibility is to add a test that calls `forward_mut` multiple times and then a final `forward` call that uses the running statistics.

src/nn/batch_norm.rs
```diff
@@ -0,0 +1,193 @@
...
+        let x = div(sub(x.duplicate(), &self.running_mean.duplicate().broadcast2()), &sd.broadcast2());
+        add(self.beta.duplicate().broadcast2(), &mul(self.gamma.duplicate().broadcast2(), &x))
+    }
+}
+
+
+impl<const B: usize, const C: usize> Module<Tensor2D<B, C, OwnedTape>> for BatchNorm1D<C> {
+    type Output = Tensor2D<B, C, OwnedTape>;
+
+    // TODO Is this the right thing to do? The input has an OwnedTape
```
coreylowman, line 95: yeah i think this is a case that isn't possible in pytorch. I think it should stick to the `&self` contract, so it shouldn't change if there's no mut, and then use running statistics.

src/nn/batch_norm.rs
```diff
@@ -0,0 +1,193 @@
...
+    // so statistcs needs to be tracked, but forward is being called
+    // instead of forward_mut, should we not allow users to call forward here?
```
coreylowman, line 97: Yeah it definitely is a bit confusing for users (& implementors)! Not sure what the best thing to do here is tbh.

I think it'd make more sense with just the following combinations allowed:
- forward_mut & OwnedTape
- forward & NoneTape

So maybe a separate trait with forward_mut would make sense? Then all modules could only impl Module with NoneTapes and ModuleMut with OwnedTapes?

vikigenius: Yeah, actually that does make sense with the context. We can reserve Module trait with forward for just NoneTape and a ModuleMut with forward_mut that only is for OwnedTape.

But that means we lose that default behavior of `forward_mut` calling `forward`. And also the genericness of forward over Tape.

Without this we will have to provide actually different implementations for `forward` and `forward_mut` for every module. Seems a bit tedious for modules that do not require different behavior w.r.t Tape

coreylowman: Hmm yeah I was originally thinking we could do something like `trait ModuleMut<Input>: Module<Input>`, but if the tapes are different in the inputs then we actually couldn't do that.

I think it also just moves the subtlety elsewhere. In dfdx at least we could enforce that Module is only for NoneTape and ModuleMut is for OwnedTape, but other libraries using dfdx wouldn't have the restriction. And I 100% agree that having two separate impls for forward would be annoying, and probably confusing for beginners.

Assuming we keep forward_mut in Module for now, do you think there is actually a use case where a BatchNorm is in your network, and you actually do want to just normalize with OwnedTape instead of tracking stats? If there isn't it may be best to just `panic!()` in that function:
1. Users would immediately know they have to change something
2. If people actually want to use that version they would raise an issue since panic makes it impossible to use
3. It avoids silent errors

vikigenius: I don't think there is a use for not tracking stats when training. But there could be that we haven't thought about it.
Pytorch does have an option to not track running stats.

Thinking of panic, I think it's a good idea, but we should also keep in mind the cons of the panic approach. People who do some long preprocessing and then start training would get frustrated that their code is panicking. Afaik, there is no `no_panic` attribute in rust to guarantee no panics, so they won't know at compile time that this is incorrect.

We could provide a warning instead, but that could actually end up being worse if it was non intentional and people missed the warning.

Overall, I think we can just panic for now, and if people find a use case for it, we can switch to a warning later.

coreylowman: @vikigenius I'm going back on what I previously said and I think having a separate ModuleMut trait is the way to go. Have a PR here: https://github.com/coreylowman/dfdx/pull/181.

One of the design goals for dfdx is that it won't compile if you're doing something wrong. Panicing at runtime in this weird case doesn't fit with that. _Maybe_ if there was a way to produce a compile warning for specific cases that would work, but AFAIK that isn't possible.

The separate traits provide
1. finer control over types of inputs & mutability 
2. actual compile time errors that show the issue (see PR)

Re the extra impl:
1. It's still straightforward to add a blanket default impl (see PR)
2. This isn't necessarily _required_ by anything. It's something dfdx::nn will adopt, but if someone wants to write their own nn layer, they don't have to do it the same way.

Overall I think very explicit double impls is worth the added code.

vikigenius: Ok, I don't have any issues with this and this seems reasonable. Coming back to the PR, is the idea (now that we have two traits) that when training users should call `forward_mut` and during inference they just call forward? 

We can add a `Module` implementation here for BatchNorm that requires `NoneTape` and a `ModuleMut` implementation that requries ownedtape to enforce this.

vikigenius:
Also, I have been pretty busy recently, so I would probably wait for https://github.com/coreylowman/dfdx/issues/163 before taking this on again.

src/nn/batch_norm.rs
coreylowman: Yep exactly!

PR #173, opened 2022-08-30
Transformers refactor

### LINKED ISSUE #122: Refactor unit tests with large arrays - move to .npy files in the repo
Unit tests should be easy to scan. large arrays make this almost impossible and add a lot of bloat.

_Originally posted by @coreylowman in https://github.com/coreylowman/dfdx/pull/120#discussion_r929261072_

### LINKED ISSUE #158: MultiHeadAttention needs to permute before reshape & after final matul
i.e.
```rust
        let tokens: Tensor4D<B, H, S, { V / H }, TAPE> = matmul(token_weights, &values);
        let tokens: Tensor3D<B, S, V, TAPE> = tokens.reshape();
```
should be something like:
```rust
        let tokens: Tensor4D<B, H, S, { V / H }, TAPE> = matmul(token_weights, &values);
        let tokens: Tensor4D<B, S, H, { V / H}, TAPE> = tokens.permute_axes::<0, 2, 1, 3>();
        let tokens: Tensor3D<B, S, V, TAPE> = tokens.reshape();
```

Depends on #108 

You can see this happen in [minGPT here](https://github.com/karpathy/minGPT/blob/master/mingpt/model.py#L66):
```python
        y = att @ v # (B, nh, T, T) x (B, nh, T, hs) -> (B, nh, T, hs)
        y = y.transpose(1, 2).contiguous().view(B, T, C) # re-assemble all head outputs side by side
```

### LINKED ISSUE #170: Can MHA Module impls be combined?
If batched & unbatched impls for 3 tuples are provided with the 2 different sequence lengths, I think that is all that is required. Additionally could just make unbatched call batched with broadcasted & reshaped inputs.

### LINKED ISSUE #171: MultiHeadAttention dividing by wrong scalar value
should be `(M // H).sqrt()`. Paper says it's `sqrt(d_k)`, but apparently d_k here is per single head attention, which means it's `M // H`

### LINKED ISSUE #172: Repeated should use vec instead of array for smaller memory footprint

coreylowman:
Closes #158
Closes #122 
Closes #170 
Closes #171
Closes #172

This turned into a way bigger refactor than I was expecting at first. Summary of changes:

1. Repeated now uses Vec instead of array (#172)
2. MultiHeadAttention changes
    1. Removed `N` param
    2. Make K & V default to M
    3. Rename M -> EMBED_DIM, H -> NUM_HEADS, K -> K_DIM, V -> V_DIM
    4. impl `SaveToNpz` and `LoadFromNpz`
    5. Unify the same seq length impls and the diff seq length impls into one. Now there are only 2 impls, 1 for batched and 1 for unbatched. Both of the impls support different seq lengths, which by def supports the same seq length. (#170)
    6. Both Module impls now take a 3-tuple of q/k/v (#170)
    7. Add appropriate calls to permute_axes (#158)
    8. Fix scalar value (#171)
3. TransformerEncoderBlock/TransformerDecoderBlock
    1. Both of these now impl Module generically, so there is only 1 impl module instead of 2
    2. Updated to use new generic params
    3. Updated to use 3-tuple input to MHA
4. TransformerEncoder/TransformerDecoder
    1. Both of these now impl Module generically, so there is only 1 impl module instead of 2
    2. Updated to use new generic params
    3. Updated to use 3-tuple input to MHA
5. Testing changes (for MHA, TransformerEncoderBlock, and TransformerDecoderBlock)
    1. They all randomly initialize all of their parameters, and their input values
    4. Only the expected output is specified. And this value is generated from passing the exact same random parameter values through the corresponding pytorch model


coreylowman:
FYA @jafioti. the addition of permute_axes made transformers implementation look a lot closer to other ones I've seen. and some additions of `Module<>` into where clauses helped reduce number of impls

Issue #142, opened 2022-08-17
Add MutModule with `&mut self` in forward


coreylowman:
As discussed in #78, there needs to be a way for modules that need to update themselves to have `&mut self` in forward.

This would impact batch norm and also dropout.

Basically all modules would need to either:
1. implement both Module and MutModule
2. just implement MutModule

Then a model with no only MutModules (e.g. `(Linear, ReLU, Linear)`) can have forward called without forward.

However when you add a only MutModule (Dropout/BatchNorm), then your model has to be mutable to call forward.

coreylowman:
An example implementation of this can be seen here: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b95daf1c648916fe4991c4de7afbe05c

```rust
    type Model = (Affine, Dropout);
    let model: Model = (
        Affine { w: 2.0, b: 0.5 },
        Dropout {
            rng: StdRng::seed_from_u64(0),
        },
    );
    let y = model.forward(1.0);
```

results in the error:
```
error[[E0596]](https://doc.rust-lang.org/stable/error-index.html#E0596): cannot borrow `model` as mutable, as it is not declared as mutable
  --> src/main.rs:76:13
   |
70 |     let model: Model = (
   |         ----- help: consider changing this to be mutable: `mut model`
...
76 |     let y = model.forward(1.0);
   |             ^^^^^^^^^^^^^^^^^^ cannot borrow as mutable
```

which goes away if you change model to mut. However if you keep model as non-mut and change the Dropout to a sescond Affine, it will compile fine.

coreylowman:
Downsides of this approach are that now every module under nn has to also implement MutModule. But that should be pretty simple like this:

```rust
impl<I> MutModule<I> for <WHATEVER STRUCT>
where
    Self: Module<I>
{
    type Output = <Self as Module<I>>::Output;
    fn forward(&mut self, x: I) -> Self::Output {
        <Self as Module<I>>::forward(self, x)
    }
}
```

coreylowman:
@vikigenius thoughts on this approach instead of just changing `Module::forward` to be `&mut`

vikigenius:
@coreylowman can you clarify this part `Then a model with no only MutModules (e.g. (Linear, ReLU, Linear)) can have forward called without forward.`

If every module has to implement `MutModule` anyway then I am not seeing the benefit to having a separate immutable `Module` implementation

coreylowman:
[...]

Issue #78, opened 2022-07-14
Add BatchNorm1D


coreylowman:
mean & variance should only be updated if the tape parameter actuall owns the tape. ie. `if TAPE::OWNS_TAPE`

The paper is: https://arxiv.org/abs/1502.03167

More details can be found at pytorch documentation: https://pytorch.org/docs/stable/generated/torch.nn.BatchNorm1d.html

vikigenius:
I was trying to implement this but am running into the following issue. We need to keep track of the running mean and running variance.

But the Module trait only allows for `fn forward(&self, input: Input) -> Self::Output;`

Shouldn't we need `fn forward(&mut self, input: Input) -> Self::Output;` if we want to update the running stats ? Or is there a way to avoid this that I am missing ?

coreylowman:
Good call out. The absolute easiest option would be wrapping them in `RefCell` (like I did with the rng in Dropout/DropoutOneIn).  However now that there are at least 2 modules that would need a mut forward I'd want to think about how to actually support that.

Do you know if there are any other module types that would need `&mut self` in forward? I think a number of the norm layers require tracking running stats right? At least InstanceNorm does off the top of my head.

Changing all forwards to be `&mut` may be overkill from an api perspective since there aren't that many modules that require it, and I like how you can run forward on non-mutable models atm. But it'd be nice to make it clear when using batch norms that the module needs to be mut.

vikigenius:
Yes, there is definitely InstanceNorm that would require it. But even apart from that if we ever get to the point where we start doing things like tracking Model metrics etc. we would need to allow forward to take a `&mut self` to update those metrics

vikigenius:
Another possible decision to make is should we implement

```rust
impl<H: Tape, const M: usize> Module<Tensor1D<M, H>> for BatchNorm1D<M> 
```

Unlike layer norm batch norm is supposed to average across the batch dimension. When you have a 1D tensor is batch_size = 1? Or do you consider it as a batch of scalars. I am more inclined to treat it as a batch size of 1 and not implement it because the batch variance would always be zero in such  a case

coreylowman:
Oh yeah this is gonna be subtle. Ran into something similar with Flatten (now called FlattenImage), cause for a 3d tensor it is most likely a single image rather than batches of 2d tensors. I am now seeing that pytorch BatchNorm1d only works for 2d & 3d tensors, and BatchNorm2d only works for 3d & 4d tensors, which addresses all of that.

[...]

PR #181, opened 2022-09-07
Add `trait ModuleMut`

### LINKED ISSUE #149: Update Dropout and DropoutOnIn to use Module::forward_mut
They should have the following behavior:

1. `Module::forward()` do nothing
2. `Module::forward_mut()` call dropout (i.e. if there is a tape, then it will modify, otherwise it will do nothing)

coreylowman:
Related to #78 #153 
Closes #149 

Changes:
1. Adds `trait ModuleMut` which has `fn forward_mut(&mut self, input: Input)`.
2. Removes ResetParams and CanUpdateWithGradients from requirements of Module
3. impl ModuleMut for Dropout structs based on #149 
4. Blanket impl ModuleMut for existing modules

This approach was previously discussed in #142 as an alternative to adding forward_mut to Module. However as we've seen in #153, simply adding forward_mut to Module allows some confusing permutations that compile but probably shouldn't. Instead of allowing weird combinations of things to happen, instead dfdx should just not compile.

Thoughts:
1. Enables module developers finer control over when self is mutably borrowed and over what inputs. This is a double edge sword because obviously this requires library developers to implement a second module trait. However as shown here it is easy to do a blanket impl for anything that impls Module.
5. Enables library developers to specify module inputs as either ModuleMut for training or Module for inference:
```rust
fn training<T, M: ModuleMut<T>>(module: &mut M) {
    ...
}

fn inference<T, M: Module<T>>(module: &M) {
    ...
}
```
6. Enables straightforward compile errors. Here is an example with the new dropout where the user calls `forward()` instead of `forward_mut()`

```rust
use dfdx::prelude::*;
fn main() {
    let mut m: (Linear<3, 3>, DropoutOneIn<2>, Linear<3, 3>) = Default::default();
    let a = m.forward(Tensor1D::<3>::zeros().trace());
}
```

```bash
error[E0271]: type mismatch resolving `<dfdx::tensor::Tensor1D<3_usize, OwnedTape> as Tensor>::NoTape == dfdx::tensor::Tensor1D<3_usize, OwnedTape>`
 --> examples\mutable.rs:5:15
  |
5 |     let a = m.forward(Tensor1D::<3>::zeros().trace());
  |               ^^^^^^^ expected struct `NoneTape`, found struct `OwnedTape`
  |
  = note: expected struct `dfdx::tensor::Tensor1D<NoneTape, _>`
             found struct `dfdx::tensor::Tensor1D<OwnedTape, _>`
  = note: required because of the requirements on the impl of `dfdx::nn::Module<dfdx::tensor::Tensor1D<3_usize, OwnedTape>>` for `dfdx::nn::DropoutOneIn<2_usize>`
  = note: 1 redundant requirement hidden
  = note: required because of the requirements on the impl of `dfdx::nn::Module<dfdx::tensor::Tensor1D<3_usize, OwnedTape>>` for `(dfdx::nn::Linear<3_usize, 3_usize>, dfdx::nn::DropoutOneIn<2_usize>, dfdx::nn::Linear<3_usize, 3_usize>)`
```

### DIFF
--- a/src/nn/activations.rs
+++ b/src/nn/activations.rs
@@ -26,6 +26,16 @@ macro_rules! activation_impls {
                 $func_name(input)
             }
         }
+
+        impl<T> ModuleMut<T> for $struct_name
+        where
+            Self: Module<T>,
+        {
+            type Output = <Self as Module<T>>::Output;
+            fn forward_mut(&mut self, input: T) -> Self::Output {
+                self.forward(input)
+            }
+        }
     };
 }
 
@@ -61,6 +71,16 @@ impl<T: Reduce1<-1>> Module<T> for Softmax {
     type Output = T;
     fn forward(&self, input: T) -> Self::Output {
         softmax(input)
+    }
+}
+
+impl<T> ModuleMut<T> for Softmax
+where
+    Self: Module<T>,
+{
+    type Output = <Self as Module<T>>::Output;
+    fn forward_mut(&mut self, input: T) -> Self::Output {
+        self.forward(input)
     }
 }
 
--- a/src/nn/conv.rs
+++ b/src/nn/conv.rs
@@ -176,6 +176,24 @@ where
     }
 }
 
+impl<
+        T,
+        const IN_CHAN: usize,
+        const OUT_CHAN: usize,
+        const KERNEL_SIZE: usize,
+        const STRIDE: usize,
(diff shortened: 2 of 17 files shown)


coreylowman:
Closing this one as this will be staged into multiple PRs

PR #182, opened 2022-09-09
Add Batched Select for devices and tensor_ops

### LINKED ISSUE #177: Add batched select
This is necessary for #121, as batches of indices are often used with transformers

coreylowman:
Closes #177 

Unblocks #121 

Batched select needs the source type to be broadcasted one level. It was actually impossible to specify this with the macro and axis based select previously used in device. This is because broadcasted select doesn't select from an axis at first, so you can't specify a single thing for it to select from.

Summary:
- Reworks the devices implementation to use a recursive trait implementation using recursive types to deconflict implementation.
    - These are controlled via unit structs called "Modes"
- Adds (private) generic select function that uses select modes to control behavior
- Rework existing tensor impls to use modes
- Adds a SelectBatchAx0 trait with select_batch function for 1d/2d/3d tensors

### DIFF
--- a/src/devices/select.rs
+++ b/src/devices/select.rs
@@ -19,16 +19,16 @@
 use super::{Cpu, ForEachElement};
 use crate::arrays::CountElements;
 
-pub mod modes {
+pub mod select_modes {
     pub struct Index;
     pub struct Recurse<const N: usize>;
     pub struct Broadcast<const N: usize>;
 }
 
-use modes::*;
+use select_modes::*;
 
 pub trait DeviceSelect<T, R, Mode> {
-    type Indices;
+    type Indices: Clone;
 
     /// Equivalent to psuedocode `out = inp[indices]`
     fn select_axis(inp: &T, indices: &Self::Indices, out: &mut R);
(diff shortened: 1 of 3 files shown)


src/devices/select.rs
```diff
@@ -18,90 +18,101 @@
 
 use super::{Cpu, ForEachElement};
 use crate::arrays::CountElements;
+use std::marker::PhantomData;
+
+pub(crate) struct Idx;
```
coreylowman, line 23: Update the module docstring and add small docstrings to these

src/devices/select.rs
```diff
@@ -18,90 +18,101 @@
...
+pub(crate) struct Rec<M>(PhantomData<*const M>);
```
coreylowman, line 24: Rename Recurse

src/devices/select.rs
```diff
@@ -18,90 +18,101 @@
...
+pub(crate) struct Bcst<M>(PhantomData<*const M>);
```
coreylowman, line 25: Rename Broadcast

src/devices/select.rs
```diff
@@ -18,90 +18,101 @@
...
+
+pub(crate) type SelectAx0 = Idx;
+pub(crate) type SelectAx1 = Rec<SelectAx0>;
+pub(crate) type SelectAx2 = Rec<SelectAx1>;
+pub(crate) type SelectAx3 = Rec<SelectAx2>;
+pub(crate) type BSelectAx0 = Bcst<Idx>;
```
coreylowman, line 31: Rename BroadcastSelectAx0

PR #236, opened 2022-10-11
Change tensors to use Arc so they implement Send and Sync

caelunshun:
Currently, tensors use `Rc` to allow cloning the tensor without reallocating the data. However, `Rc` is neither `Send` or `Sync`, which prevents tensors and everything that contain them (most notably, Modules) from being moved to another thread.

For example, it is currently impossible to share a network across threads except via suboptimal workarounds like serializing the network to bytes on one thread and deserializing it on the other.

By using `Arc`, tensors now implement `Send` and `Sync`. The cost of using atomic operations should be negligible.


coreylowman:
Hey thanks for this PR, this is a great minimal diff to look at and really big improvements to thread sharing! 😁 🚀

Issue #145, opened 2022-08-20
Parallelize conv_batched


coreylowman:
This can be done with rayon.

Parallelizing forward is relatively straightforward, as you can just `result.par_iter_mut().zip(...).foreach()` since filters & bias are ref borrowed.

Parallelizing backward is less striaghtforward, but easiest way is to allocate a filters & bias with batch size dimension, and then par_iter_mut() on those two and img_grad. Then at the end of backward up, sum the batched filters & bias into filter_grad & bias_grad.

```rust
pub fn conv2d_par_batched<
    TAPE: 'static + Tape,
    const BATCH_SIZE: usize,
    const IN_CHAN: usize,
    const OUT_CHAN: usize,
    const KERNEL: usize,
    const STRIDE: usize,
    const PADDING: usize,
    const IN_HEIGHT: usize,
    const IN_WIDTH: usize,
>(
    x: Tensor4D<BATCH_SIZE, IN_CHAN, IN_HEIGHT, IN_WIDTH, TAPE>,
    filters: &Tensor4D<OUT_CHAN, IN_CHAN, KERNEL, KERNEL>,
    bias: &Tensor1D<OUT_CHAN>,
) -> Tensor4D<
    BATCH_SIZE,
    OUT_CHAN,
    { (IN_HEIGHT + 2 * PADDING - KERNEL) / STRIDE + 1 },
    { (IN_WIDTH + 2 * PADDING - KERNEL) / STRIDE + 1 },
    TAPE,
> {
    let mut result = Tensor4D::zeros();
    let f_data = filters.data();
    let b_data = bias.data();
    result
        .mut_data()
        .par_iter_mut()
        .zip(x.data().par_iter())
        .for_each(|(r_i, x_i)| {
            conv_forward::<
                IN_CHAN,
                OUT_CHAN,
                KERNEL,
                STRIDE,
                PADDING,
                IN_HEIGHT,
                IN_WIDTH,
                { (IN_HEIGHT + 2 * PADDING - KERNEL) / STRIDE + 1 },
                { (IN_WIDTH + 2 * PADDING - KERNEL) / STRIDE + 1 },
            >(x_i, f_data, b_data, r_i);
        });

    let f = filters.clone();

    let (x, mut tape) = x.split_tape();
    let phantom_filters = filters.phantom();
    let phantom_bias = bias.phantom();
    let phantom_result = result.phantom();
    tape.add_backward_op(move |grads| {
[...]

PR #237, opened 2022-10-12
Replacing conv2d implementation with matmuls

coreylowman:
This addresses some slowness in conv2d implementation by leveraging matrix multiplication algorithms.

Here are some numbers from my local benchmarks with `Conv2D<128, 256, 4>` on `Tensor4D<64, 128, 28, 28>`:

| | pytorch| pytorch single thread* | matrixmultiply | mkl-dynamic-seq | mkl-dynamic-iomp | old naive approach |
|---| ---|--- | --- | --- | --- | --- |
| forward | 163ms | 400ms | 1597ms | 524ms | 382ms | 2000ms |
| backward | 293ms | 795ms | 4100ms | 1245ms | 876ms | 7714ms |

* pytorch has num_threads and num_interop_threads set to 1. Parallel batched convs will be addressed by #145 

There is probably still more work that can be done as far as reducing allocations/copying when creating the patches buffers used below.

### DIFF
--- a/src/devices/conv.rs
+++ b/src/devices/conv.rs
@@ -1,4 +1,8 @@
-use super::{AllocateZeros, Cpu, FillElements, MatMul};
+use super::{AllocateZeros, Cpu, FillElements};
+#[cfg(feature = "cblas")]
+use cblas_sys::{
+    cblas_sgemm as sgemm, CblasNoTrans as NoTr, CblasRowMajor as RowMajor, CblasTrans as Tr,
+};
 
 /// **Requires nightly** 2d convolution with stride and padding specified at trait level.
 ///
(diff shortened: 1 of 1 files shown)


seddonm1:
Did you run the benchmarks for the previous 'naive' implementation?

coreylowman:
> Did you run the benchmarks for the previous 'naive' implementation?

Just added - they are really slow lol. Didn't scale up the number of channels enough in my original benchmarks of convs, but it was a really simple algorithm to debug and test. All the unit tests made it really easy to verify that this new implementation worked! 😁

src/devices/conv.rs
```diff
@@ -48,28 +55,46 @@ impl<const S: usize, const P: usize> DeviceConv2D<S, P> for Cpu {
...
+            matrixmultiply::sgemm(
+                m, k, n, 1.0, a, k as isize, 1, b, n as isize, 1, 1.0, c, n as isize, 1,
+            )
+        }
+
+        #[cfg(feature = "cblas")]
+        unsafe {
+            let (m, n, k) = (m as libc::c_int, n as libc::c_int, k as libc::c_int);
+            sgemm(RowMajor, NoTr, NoTr, m, n, k, 1.0, a, k, b, n, 1.0, c, n)
+        }
```
coreylowman, line 96: Noting that this is copied from the matmul implementation - using Cpu::mm requires casting the inputs to 2d arrays, which in turn requires generic_const_expr bounds added to the trait, which I wanted to avoid.

coreylowman:
> Damn, I'd hoped for some LLVM intrinsics magic that kept the naive way somewhat competitive! Looks like Fortran will be with us for a lot longer 😝

Hah yeah I was hoping that too. I'm pretty sure it was auto vectorizing the forward at least, but I think all the matmul stuff is suuuuper cache optimized.

PR #244, opened 2022-10-15
Adding `#![no_std]` support via `no_std_compat`

### LINKED ISSUE #201: `no_std` support
First asked in #196 

- [x] use `allocations` for `Rc` and allocating
- [x] use `collections` for hashmap and vec
- [x] add `std` feature
- [x] only depend on zip when std is enabled
- [x] only include serialization & numpy with `std` feature

coreylowman:
Closes #201

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ keywords = [
 features = ["nightly", "numpy"]
 
 [dependencies]
+no-std-compat = { version = "0.4.1", features = [ "alloc", "compat_hash" ] }
 rand = { version = "0.8.5", features = ["std_rng"] }
 rand_distr = { version = "0.4.3", features = [] }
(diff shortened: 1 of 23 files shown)


Cargo.toml
```diff
@@ -24,6 +24,7 @@ keywords = [
 features = ["nightly", "numpy"]
 
 [dependencies]
+no-std-compat = { version = "0.4.1", features = [ "alloc", "compat_hash" ] }
```
coreylowman, line 27: @antimora Is no_std still useful even if alloc will be required?

antimora: @coreylowman Awesome! Thank you for adding the support. Yes, it still will be very useful. Sorry, I missed your message earlier. I came back to see the status of this request and you added it! Great, thanks!

PR #249, opened 2022-10-16
Clone now keeps same id, removing Tensor::duplicate

### LINKED ISSUE #248: Usage of `Clone` is confusing
This weirdness comes from:
1. Cloning a Module needs to have different tensor ids for backprop to work as expected, so cloning a tensor will change the tensor's id
2. Cloning is not implemented for Tensors with an OwnedTape (why not just clone the owned tape?)
3. `.duplicate()` must be used on tensors with an OwnedTape to get a copy of it without the tape

This is weird because:
1. it may not be clear why you can clone on some tensors (NoneTape) and not others (OwnedTape)
2. Unclear when to use duplicate vs clone
3. Cloning a tensor isn't actually a clone since it changes the id

See discussion starting at https://discord.com/channels/998570811650748417/998570812590260336/1031186211643084880

coreylowman:
Closes #248

### DIFF
--- a/src/tensor_ops/utils.rs
+++ b/src/tensor_ops/utils.rs
@@ -38,7 +38,7 @@ where
 {
     T::Device::foreach_m(t.mut_data(), &mut |x| *x = f(x)); // clones if there is more than 1 reference to t
     let (t, mut tape) = t.split_tape();
-    let result = t.clone(); // will always a new reference to t, not start a new one
+    let result = t.clone().detach(); // result will have a new id but same data as t
     let phantom_result = result.phantom();
     tape.add_backward_op(move |grads| {
         let (t_grad, result_grad) = grads.mut_and_ref(&t, &phantom_result);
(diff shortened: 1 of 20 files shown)


src/tensor_ops/utils.rs
```diff
@@ -38,7 +38,10 @@ where
 {
     T::Device::foreach_m(t.mut_data(), &mut |x| *x = f(x)); // clones if there is more than 1 reference to t
     let (t, mut tape) = t.split_tape();
-    let result = t.clone(); // will always a new reference to t, not start a new one
+
+    let mut result: T::NoTape = TensorCreator::zeros();
+    result.mut_data().clone_from(t.data());
```
coreylowman, line 43: Is it possible to do this with re-allocating every time? this function was first introduced because it could reduce allocations for certain things, like softmax:
```rust
x.log_softmax().exp()
```

Here only 1 thing will be referencing `x.log_softmax()`, so the reference count will be 1. That means we can just re-use the memory instead of allocating new data.

PR #255, opened 2022-10-17
Make ZeroElements unsafe, and AllocateZeros sound

### LINKED ISSUE #252: Cpu::zeros::<T> is unsound for some T
There's a [comment in the implementation](https://github.com/coreylowman/dfdx/blob/main/src/devices/allocate.rs#L14) asking "TODO is this function safe for any T?". It isn't, various types are defined to not take on various values (rust calls these niches), and forcing them to do so is undefined behavior.

Examples include `std::num::NonZeroU32` and friends, references, any enum with no variant that is tagged as 0, etc.

Below is an example where miri will catch undefined behavior and complain ("encounted 0, but expected something greater or equal to 1").

```rust
use dfdx::{
    arrays::CountElements,
    devices::AllocateZeros,
};

#[repr(transparent)]
#[derive(Clone, Copy)]
struct Illegal {
    inner: std::num::NonZeroU32
}

impl CountElements for Illegal {
    type Dtype = Self;
    const NUM_ELEMENTS: usize = 1;
    const NUM_BYTES: usize = 4;

    fn ref_first_elem(&self) -> &Self::Dtype {
        todo!()
    }

    fn mut_first_elem(&mut self) -> &mut Self::Dtype {
        todo!()
    }
}

fn main() {
    dfdx::devices::Cpu::zeros::<Illegal>().inner;
}
```

gmorenz:
fixes #252 

Takes the approach I just mentioned in #252 of piggy-backing on `ZeroElements` instead of introducing another trait. There is a trade-off here  that this makes `ZeroElements` less flexible, but realistically I think every dtype ends up needing to be `ValidAsAllZeroBits` anyways, so I don't think the flexibility was likely to be useful

### DIFF
--- a/src/arrays.rs
+++ b/src/arrays.rs
@@ -154,16 +154,16 @@ impl<const M: usize, const N: usize, const O: usize, const P: usize> HasLastAxis
     const SIZE: usize = P;
 }
 
-/// Something that has compile time known zero values.
-pub trait ZeroElements {
+/// Something that has a valid 0 value represented by all 0 bits.
+pub unsafe trait ZeroElements {
     const ZEROS: Self;
 }
 
-impl ZeroElements for f32 {
+unsafe impl ZeroElements for f32 {
     const ZEROS: Self = 0.0;
 }
 
-impl<T: ZeroElements, const M: usize> ZeroElements for [T; M] {
+unsafe impl<T: ZeroElements, const M: usize> ZeroElements for [T; M] {
     const ZEROS: Self = [T::ZEROS; M];
 }
 
--- a/src/devices/allocate.rs
+++ b/src/devices/allocate.rs
@@ -1,21 +1,21 @@
 use super::Cpu;
-use crate::arrays::CountElements;
+use crate::arrays::{CountElements, ZeroElements};
 use std::alloc::{alloc_zeroed, Layout};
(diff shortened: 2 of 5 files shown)


src/arrays.rs
```diff
@@ -154,16 +154,16 @@ impl<const M: usize, const N: usize, const O: usize, const P: usize> HasLastAxis
     const SIZE: usize = P;
 }
 
-/// Something that has compile time known zero values.
-pub trait ZeroElements {
+/// Something that has a valid 0 value represented by all 0 bits.
+pub unsafe trait ZeroElements {
```
coreylowman, line 158: I think I still prefer having a separate trait rather than re-use ZeroElements mainly because of naming. Would rather someone just look at the type and know it has to be valid as all 0 bits than read doc string.

ZeroElements is just intended to expose a const value that has everything 0'd out, which doesn't necessarily have to do with 0 bits (e.g. you mentioned `std::num::NonZeroU32` in the issue, which seems like a `impl ZeroElements` but NOT `impl ValidAsAllZeroBits`).

coreylowman (changes requested):
Looks good overall, and nice little diff 🚀 Let's make a separate trait instead of using ZeroElements and then we can get this merged!

PR #256, opened 2022-10-18
AddInto

Dimev:
AddInto, to allow easier multiple inputs for networks

### DIFF
--- a/src/nn/add_into.rs
+++ b/src/nn/add_into.rs
@@ -66,34 +66,29 @@ macro_rules! tuple_impls {
             $headin: Tensor<Dtype = f32>,
             $($tailsin: Tensor<Dtype = f32>,)+
             $head: Module<$headin, Output = Output>,
-            $($tails: Module<$tailsin, Output = Output>,)+
+            $($tails: Module<$tailsin>,)+
         > Module<($headin, $($tailsin,)+)> for AddInto<($head, $($tails,)+)> {
             type Output = Output;
             
+            #[allow(non_snake_case)]
             fn forward(&self, x: ($headin, $($tailsin,)+)) -> Self::Output {
-                
-                // split tape on input
-                let (head, tape) = x.0.split_tape();
-                
-                // forward all modules on the tape
-                
-                // add them all together
-                
-                // modules
-                let ($head, $($tails),+) = self.0;
-                let head_module = $head;
-                let tail_module = AddInto(($($tails),+,));
                 
                 // inputs
                 let ($head, $($tails),+) = x;
-                let head_in = $head;
-                let tails_in = ($($tails),+,);
+                let ($head, tape) = $head.split_tape();
+                
+                // layers
+                let ($headin, $($tailsin),+) = &self.0;
                 
                 // forward
-                let head_x = head_module.forward(head_in);
-                let tails_x = tail_module.forward(tails_in);
+                let ($head, $($tails),+) = ($headin.forward($head.put_tape(tape)), $($tailsin.forward($tails)),+);
                 
-                add(tails_x, &head_x)
+                // add together
+                $(
+                    let $head = add($head, &$tails);
+                )+
+                
+                $head
             }
         }
     }
--- a/src/nn/add_into.rs
+++ b/src/nn/add_into.rs
@@ -1,0 +1,54 @@
+use crate::gradients::{CanUpdateWithGradients, GradientProvider, UnusedTensors};
+use crate::prelude::*;
+
+/// Add inputs together into a single tensor. `T` should be a tuple
+//// where every element of the tuple has the same output type
+///
+/// This provides a utility for networks where multiple inputs are needed
+///
+/// # Generics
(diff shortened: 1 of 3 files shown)


Dimev:
adds `AddInto<T>`, the opposite of `SplitInto<T>`, allowing multiple inputs to a network by adding the inputs together

src/nn/add_into.rs
```diff
@@ -0,0 +1,238 @@
...
+            fn forward_mut(&mut self, x: ($headin, $($tailsin,)+)) -> Self::Output {
+
+                // inputs
+                let ($head, $($tails),+) = x;
+
+                // layers
+                let ($headin, $($tailsin),+) = &self.0;
+
+                // forward
+                let ($head, $($tails),+) = ($headin.forward($head), $($tailsin.forward($tails)),+);
```
coreylowman, line 86:
```suggestion
                let ($head, $($tails),+) = ($headin.forward_mut($head), $($tailsin.forward_mut($tails)),+);
```

src/nn/add_into.rs
```diff
@@ -0,0 +1,238 @@
...
+            }
+        }
+
+
+        impl<
+            Output: Tensor<Dtype = f32>,
+            $headin: Tensor<Dtype = f32>,
+            $($tailsin: Tensor<Dtype = f32>,)+
+            $head: Module<$headin, Output = Output>,
+            $($tails: Module<$tailsin, Output = Output>,)+
```
coreylowman, line 72:
```suggestion
            $head: ModuleMut<$headin, Output = Output>,
            $($tails: ModuleMut<$tailsin, Output = Output>,)+
```

src/nn/add_into.rs
```diff
@@ -0,0 +1,238 @@
...
+    #[test]
+    fn test_add_into_2() {
+        type Model = AddInto<(Linear<2, 5>, Linear<3, 5>)>;
+        let m: Model = Default::default();
+        let _: Tensor1D<5, OwnedTape> =
+            m.forward((Tensor1D::zeros().traced(), Tensor1D::zeros().traced()));
+        let _: Tensor2D<3, 5, OwnedTape> = m.forward((
+            Tensor2D::<3, 2>::zeros().traced(),
+            Tensor2D::<3, 3>::zeros().traced(),
+        ));
```
coreylowman, line 118: Side comment, but I think with the way this is set up you actually only need the tape on the first input if I'm understanding the forward correctly.

coreylowman (changes requested):
Looks good so far! Pretty easy to understand the macros too. Can you make the following changes:
1. a couple small fixes for ModuleMut (as I suggested in the comments)
2. can you add serialization for this into `src/nn/npz_impls.rs`? It should be relatively easy since you can just use the sub modules implementation for serialization

src/nn/add_into.rs
Dimev: Yeah, wasn't really sure about how to handle this here, as I'm not super familliar with how tapes work.

Dimev:
Alright, will make the changes when I have the time!

Dimev:
for npz saving/loading I can probably just copy the implementation for SplitInto as it's effectively the same struct, just different functionality

src/nn/add_into.rs
Dimev: Fixed

src/nn/add_into.rs
Dimev: Fixed

src/nn/add_into.rs
Dimev: basically the forward expands to
```rust
// tuple, so we get all layers
let (headin, ain, bin, cin) = &self.0;

// forward
let (head, a, b, c) = (headin.forward(), ain.forward(), bin.forward(), cin.forward());

// add all of them together recursively
let head = add(head, a);
let head = add(head, b);
let head = add(head, c);

head
```

Dimev: hmm, looking at the add and binary_map functions it seems those merge the tapes, so maybe it's not needed? 

Both this and doing the with_new_tape on the input tails before doing forward passes the test_missing_gradients test

coreylowman: Ahh yeah the expansion is helpful, so the ain/bin/cin forwards aren't recorded unless their inputs have tapes. Should probably think about whether this should require that all inputs have tapes or not - ideally this could work in both cases. That can probably be done in a separate follow up PR though, I think expecting all of them to have tapes at first is reasonable enough

Dimev: Yeah makes sense. 
So should I change it to have it expect tapes for all the inputs?

coreylowman: Nah I think it's good how it is, I'll just make an issue based on this thread for tracking purposes. I think this is good to merge though!

coreylowman (approved):
🚀 thanks for the contribution!

Dimev:
Noice, thanks for making such a cool crate for neural nets!

Issue #99, opened 2022-07-20
Add Weight Decay to optimizers


coreylowman:
- [x] Sgd `weight_decay: Option<f32>`
- [ ] RMSprop `weight_decay: Option<f32>`
- [ ] Adam/AdamW: `weight_decay: Option<AdamDecay>`, where `enum AdamDecay { Vanilla(f32), AdamW(f32) }`.

(maybe use different terms for `AdamDecay`).

cBournhonesque:
Hi, I wanted to work on this issue but I think I need some guidance.

1. If I understand correctly, the gradients returned by [this function](https://github.com/coreylowman/dfdx/blob/main/src/optim/sgd.rs#L120) are separate from the gradients present in the tensors' tape, and are safe to be modified? i.e. modifying the gradients inside this function won't modify the gradients associated with each parameter weight, they only represent the 'update' to be passed to each tensor during the optimization step.

2. Provided 1. is true and I can modify `*g` inside that function directly, the only thing I need to do is to add some code like
```
if let Some(wd) = self.cfg.weight_decay {
            P::Device::foreach_m(g_t.as_mut(), &mut |g| *g += wd * param_weights )
}
```
above [this line](https://github.com/coreylowman/dfdx/blob/main/src/optim/sgd.rs#L142) ?

3. The other issue is that I would need to have access to the param weights inside the sgd `gradient` function. It seems like there needs to be a function `weights` in the `Module` trait that gives me access to all the param weights of a module? (an object similar to `Gradients`, something like a HashMap<id, Array>)

coreylowman:
1. Yeah they are separate from gradients inside the tape. You can kind of think about optimizers as a mapping function that take the raw gradient from the tape, apply some functions to it, and then return it so it can be used to update the tensor.
2. Yep! But double check which line exactly should get the weight decay update. The pytorch docs have good descriptions for the optimizers. For example in [SGD](https://pytorch.org/docs/stable/generated/torch.optim.SGD.html) they show the weight decay operation happening before the momentum updates
3. The `p: &P` parameter passed into gradient function is the parameter! So you can call `p.data()` to get a reference to the array.

Also don't feel the need to add the functionality to all optimizers before opening a PR. Will be easier to review if they are added 1 at a time. Probably should've split this issue into 3 separate ones in retrospect haha

cBournhonesque:
Another thing, I'm having a bit of trouble with the tests, because of numerical instability of how floats are represented?
[...]

PR #258, opened 2022-10-18
Add weight decay to SGD

cBournhonesque:
PR to add weight_decay to SGD.

Weight_decay is added after any momentum updates, so is NOT equivalent to L2 regularization.
(this is the similar to what is done in [tensorflow](https://github.com/tensorflow/addons/blob/v0.17.0/tensorflow_addons/optimizers/weight_decay_optimizers.py#L53) or AdamW in Pytorch).

Note: I had to add `HasArrayData` to `GradientProvider` because weight_decay makes use of the array data; I hope that's ok. Is there a case where we want to distinguish between a GradientProvider that can access the array data and one that cannot?

Related issue: #99

### DIFF
--- a/src/optim/sgd.rs
+++ b/src/optim/sgd.rs
@@ -10,6 +10,11 @@ use std::{boxed::Box, marker::PhantomData};
 /// Nesterov Momentum is implemented as described in
 /// [On the importance of initialization and momentum in deep learning](https://proceedings.mlr.press/v28/sutskever13.html).
 ///
+/// Weight decay is implemented as described in
+/// [Decoupled Weight Decay Regularization](https://arxiv.org/abs/1711.05101)
+/// Note that weight decay is applied after momentum updates and is not equivalent to L2
+/// regularization when momentum is used.
+///
 /// # Example Usage
 ///
 /// Constructing using default:
@@ -26,6 +31,7 @@ use std::{boxed::Box, marker::PhantomData};
 /// let mut opt: Sgd<Model> = Sgd::new(SgdConfig {
 ///     lr: 1e-3,
 ///     momentum: Some(Momentum::Classic(0.5)),
+///     weight_decay: None,
 /// });
 /// ```
 ///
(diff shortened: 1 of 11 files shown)


src/optim/sgd.rs
```diff
@@ -244,6 +276,54 @@ mod tests {
...
+            [0.9988, 0.997, 0.995, 0.979, 0.799],
+            [0.99760115, 0.994003, 0.990005, 0.958021, 0.59820104],
+            [0.9964036, 0.991009, 0.98501503, 0.937063, 0.39760286],
+            [0.9952072, 0.988018, 0.98003, 0.9161259, 0.19720526],
+            [0.994012, 0.98502994, 0.97505, 0.8952098, -0.0029919297],
+        ];
+        for e in expected.iter() {
+            let gradients = backward((t.trace() * &rate).mean());
+            sgd.update(&mut t, gradients).expect("");
+            assert_eq!(t.data(), e);
```
coreylowman, line 299: For non-exact results you can use `crate::tests::assert_close` instead of `assert_eq!`. Have been slowly moving tests to use this instead, but all the optimizers still use assert_eq

coreylowman:
Added #260 for the HasArrayData, that can probably be a PR separate from this for easy merging.

src/optim/sgd.rs
coreylowman: Added #259 for this

src/optim/sgd.rs
```diff
@@ -240,7 +297,111 @@ mod tests {
...
+            assert_close(t.data(), e);
+        }
+
+        // Should be equivalent to l2 regularization, even with momentum
+        t = Tensor1D::ones();
+        for e in expected.iter() {
+            let normal_loss = (t.trace() * &rate).mean();
+            let (normal_loss, tape) = normal_loss.split_tape();
+            let l2_loss = mul_scalar(t.clone().put_tape(tape).powi(2).sum(), weight_decay / (2.0));
+            let loss = add(l2_loss, &normal_loss);
```
coreylowman, line 400: Nice 😁

coreylowman (approved):
Nice implementation and great unit tests, thanks for this!

Issue #95, opened 2022-07-19
Derive CanUpdateWithGradients and ResetParams


jafioti:
It seems that `CanUpdateWithGradients` is pretty simple and should be able to be derived so that each tensor in the model is updated with the grads. `ResetParams` is a little more tricky, so we could just have a default random initialization scheme (maybe just uniform 0 to 1), and then if custom initialization is desired, a manual implementation is needed.

Would be great instead of doing this:
```rust
#[derive(Default, Debug, Clone)]
pub struct Linear<const I: usize, const O: usize> {
    pub weight: Tensor2D<O, I, NoneTape>,
    pub bias: Tensor1D<O, NoneTape>,
}

impl<const I: usize, const O: usize> CanUpdateWithGradients for Linear<I, O> {
    fn update<G: GradientProvider>(&mut self, grads: &mut G) {
        self.weight.update(grads);
        self.bias.update(grads);
    }
}
```
we can just do this:
```rust
#[derive(Default, Debug, Clone, CanUpdateGradients)]
pub struct Linear<const I: usize, const O: usize> {
    pub weight: Tensor2D<O, I, NoneTape>,
    pub bias: Tensor1D<O, NoneTape>,
}
```

These changes would of course require a separate proc macros crate though, so it will be a non trivial change.

cBournhonesque:
I'd like to learn more about derive/macros, I'd be happy to work on this

cBournhonesque:
Looking at other procedural macros, it looks like I would have to change a bit the structure of the repo?
Seems like procedural macros have to be in their own crate, so maybe we need to use rust workspaces?

Similar to https://github.com/tokio-rs/tokio 

We would have:
dfdx (git repo)
- dfdx (main crate)
  - Cargo.toml
- dfdx-macros (macros crate)
  - Cargo.toml
- examples (examples crate)
- Cargo.toml

The main Cargo.toml would contain this
```
[workspace]
members = [
  "dfdx",
  "dfdx-macros",
  # Internal
  "examples",
 ]
 
Does that sound good to you?

coreylowman:
I think we can use the [root package](https://doc.rust-lang.org/cargo/reference/workspaces.html#root-package) feature of workspaces to simplify this a bit.

Specifically:
1. Add `[workspace]` at the top of `Cargo.toml`
2. Add a `dfdx-macros/` lib at the *same level* as src
3. Add `dfdx-macros = { path = "./dfdx-macros", optional = true }` under dependencies

[...]

PR #269, opened 2022-10-20
Restructure repo to use workspaces

cBournhonesque:
We'd like to create procedural macros to help reduce the amount of boilerplate code in the repo: https://github.com/coreylowman/dfdx/issues/95

Procedural macros need to be in their own crate, so we would need to restructure the repo to make use of rust workspaces, similarly to what tokio is doing. https://github.com/tokio-rs/tokio


cBournhonesque:
@coreylowman what do you think? I saw your comment in the issue that suggests using a root workspace
I like this setup because there's no need to add `--workspace` everytime.
+ some blog post was recommending it: https://matklad.github.io/2021/08/22/large-rust-workspaces.html

coreylowman:
@cBournhonesque thanks for sharing that article, it's super useful. Let me think about this for a couple days. On one hand, being consistent with other rust repos & having a standard layout is really good. On the other hand, I don't have any plans *currently* to add any more crates (than this macros one) or organize dfdx as multiple crates. The root workspace seems much simpler to me, and I'm also not sure if dfdx falls under the same category of "large workspace" that rust-analyzer does.

coreylowman:
@cBournhonesque let's just use virtual root workspace for now. This will make the diff easier to review, and I can always move to the more scalable structure later

cBournhonesque:
Sg! will close this and open another PR

PR #273, opened 2022-10-26
CanUpdateWithDerive procedural macro

cBournhonesque:
Add a CanUpdateWithGradients macro. (https://github.com/coreylowman/dfdx/issues/95)

Note:

- the types generated in the macro are written using a full path (`dfdx::gradients::GradientProvider`), so that users of the macro don't have to know a "hidden rule" of having to import `dfdx::gradients::GradientProvider` to use the macro
- to make this work, I had to add `extern crate self as dfdx;` in `main/lib.rs`: https://users.rust-lang.org/t/macro-failed-to-resolve-use-of-undeclared-crate-or-module/65009/3 
- I added `CanUpdateWithGradients` in the prelude, but it can be removed if unwanted.
- I added `dfdx-macros` as a non-optional dependency in `dfdx` since it's used to generate code in `dfdx::nn`

## Tests

- Tests pass in dfdx-macros: `cargo test`

- Can expand code in dfdx-macros: `cargo expand --test test_can_upgrade_with_gradients`

- Tests pass in dfdx: `cargo test`

- Can expand code in dfdx-macros to see that it's equivalent to the existing code: `cargo expand nn::generalized_residual`

Next steps:
1) add a `ResetParams` macro
2) adding [derive macro attributes](https://doc.rust-lang.org/stable/reference/procedural-macros.html#derive-macro-helper-attributes) so that the user can specify that some attributes shouldn't be updated. @coreylowman would this be worth it or is this overkill?

### DIFF
--- a/dfdx-macros/src/lib.rs
+++ b/dfdx-macros/src/lib.rs
@@ -1,0 +1,92 @@
+//! Macros for use with dfdx
+
+// This `extern` is required for older `rustc` versions but newer `rustc`
+// versions warn about the unused `extern crate`.
+#[allow(unused_extern_crates)]
+extern crate proc_macro;
+
+use proc_macro2::TokenStream;
+use procout::procout;
+use quote::{format_ident, quote, quote_spanned};
+use syn::spanned::Spanned;
+use syn::{DeriveInput, parse_macro_input};
+
+
+/// Implements CanUpdateWithGradients for a Module
+///
+/// ```rust
+/// use dfdx::prelude::*;
+/// use dfdx_macros::CanUpdateWithGradients;
+///
+/// #[derive(CanUpdateWithGradients)]
+/// pub struct Linear<const I: usize, const O: usize> {
+///     // Transposed weight matrix, shape (O, I)
+///     pub weight: Tensor2D<O, I>,
+///
+///     // Bias vector, shape (O, )
+///     pub bias: Tensor1D<O>,
+/// }
+/// ```
+#[proc_macro_derive(CanUpdateWithGradients)]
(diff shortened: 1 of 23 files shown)


dfdx-macros/src/lib.rs
```diff
@@ -0,0 +1,92 @@
...
+    procout(&TokenStream::from(code_block.clone()), Some(format_ident!("{}", name)), Some("macro-out"));
+
+    proc_macro::TokenStream::from(code_block)
+}
+
+fn get_updates(data: &syn::Data) -> TokenStream {
+    // Currently only works if a struct doesn't have a mix of named and unnamed fields
+
+    // TODO: have attributes to mark which fields to add grads to (see batchnorm).
+    //  maybe one attribute `nograd` and all the fields below won't have grads?
```
coreylowman, line 60: good call out - i think for first pass we can just derive everything, and if you don't want that behavior, then you'll have to impl CanUpdateWithGradients directly.

coreylowman:
@cBournhonesque yeah this looks good to me so far.
1. for importing dfdx into dfdx-macros, this is just for tests so far? what's the standard thing to do for proc macros? I'm wondering if we even need to import dfdx - can we just ensure the generated code matches some expected structure?
2. where are you trying to import it from? examples? src?

cBournhonesque:
@coreylowman any thoughts?

coreylowman:
@cBournhonesque sorry for delay - any chance you'd be willing to revisit this? Things have changed quite a bit on main to support different devices (& changed the trait name to GradientUpdate). I still think this would be great a great add, and looking at the diff it looks good from my end (after going off current main).

coreylowman:
Closing for now - feel free to reopen!

Issue #279, opened 2022-11-01
Adding initializers


cBournhonesque:
I know that we have the trait `ResetParams` for modules, and `Randomize` for tensors.

I'm trying to see how we can make them more user-friendly, by having something similar to https://keras.io/api/layers/initializers/#randomnormal-class 

1) Would it be worthwhile to have the ResetParams trait take a Distribution object that the user could provide?
So that the user has more control over how the params of the networks are reset.

2) Would it be useful to have a module `initializers` similar to Keras?
Note that some initializers make use of the shape of the module that calls them.

coreylowman:
Great question. Before ResetParams they just implemented the Randomize that tensors implement, but then you couldn't have things that depend on shape as you mentioned.

I wonder if it's possible to have some sort of fallback impl for initializers? Thinking along the lines of this:
```rust
// NOTE: use like: `StandardInitializers.reset_params(&mut model)`
pub struct StandardInitializers;
impl ResetParams<Linear<I, O>> for StandardInitializers { }
```

Then a user could override with:

```rust
pub struct MyLinearInitializers;
impl ResetParams<Linear<I, O>> for MyLinearInitializers { ... }
impl<M> ResetParams<M> for MyLinearInitializers where StandardInitializers: ResetParams<M> { ... }
```

However I think rust would error out with conflicting implementation error (since the StandardInitializers implements something for Linear<I, O> already)

coreylowman:
This *might* be do-able with TensorCollection as it is now, especially since ResetParams was moved to doing this. Check out the existing implementation of ResetParams:

```rust
struct Resetter;
impl<E: Dtype, D: DeviceStorage> TensorVisitor<E, D> for Resetter {
    type Viewer = ViewTensorMut;
    type Err = D::Err;

    fn visit<S: Shape>(
        &mut self,
        _: String,
        opts: TensorOptions<S, E, D>,
        t: &mut Tensor<S, E, D>,
    ) -> Result<(), D::Err> {
        (opts.reset)(t)
    }
}
pub trait ResetParams<E: Dtype, D: DeviceStorage>: TensorCollection<E, D> {
    fn reset_params(&mut self) {
        self.try_reset_params().unwrap();
    }
    fn try_reset_params(&mut self) -> Result<(), D::Err> {
        Self::iter_tensors(&mut RecursiveWalker {
            m: self,
            f: &mut Resetter,
            path: &mut Vec::new(),
        })
    }
}
[...]

PR #280, opened 2022-11-01
Procedural macro: reset params

cBournhonesque:
Followup PR: Must be merged after https://github.com/coreylowman/dfdx/pull/273 
Relevant Issue(s): https://github.com/coreylowman/dfdx/issues/95, https://github.com/coreylowman/dfdx/issues/279 

This PR adds a procedural macro for ResetParams.

Note:

- rand::Rng or rand_distr::StandardNormal must be imported if they are generated by the macro

- I show an example of how procedural macro attributes could work to customize the code-generation. I think this might need a bit more work; and we might want a better `Initializer` trait defined in the main crate?

### DIFF
--- a/dfdx-macros/src/reset_params.rs
+++ b/dfdx-macros/src/reset_params.rs
@@ -1,0 +1,29 @@
+// Helper functions for the ResetParams derive macro
+use darling::FromDeriveInput;
+use proc_macro2::TokenStream;
+use quote::quote;
+
+#[derive(Debug, FromDeriveInput, Default)]
+#[darling(default, attributes(reset_params), forward_attrs(allow, doc, cfg))]
+struct Opts {
+    answer: Option<i32>,
+    etc: Option<String>,
+}
+
+pub fn gen(ast: syn::DeriveInput) -> TokenStream {
+    let name = &ast.ident;
+    let opts = Opts::from_derive_input(&ast).expect("Wrong options");
+    println!("{:?}", opts);
+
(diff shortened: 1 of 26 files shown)


src/nn/transformer/mha.rs
```diff
@@ -19,7 +22,7 @@ use rand::Rng;
 /// - `MultiHeadAttention<8, 2, 6, 4>` is an attention layer with the key and value dimension different
 ///   than the embed dimension
 /// TODO: Doctests fail for some reason
-#[derive(Debug, Clone, Default)]
+#[derive(Debug, Clone, Default, CanUpdateWithGradients)]
```
coreylowman, line 25: Same question here!

dfdx-macros/src/reset_params.rs
```diff
@@ -0,0 +1,144 @@
+// Helper functions for the ResetParams derive macro
+use darling::{Error, FromDeriveInput, FromMeta, Result};
+use proc_macro2::TokenStream;
+use quote::{quote, quote_spanned};
+use syn::spanned::Spanned;
+
+#[derive(Debug)]
+enum Initializers {
```
coreylowman, line 8: I think for now we can use this ResetParams for only handling modules that just call reset on all sub modules. Things that have actual logic in reset params can just implement by hand

coreylowman (commented):
Would be great to get this in, would simplify a lot of the modules

coreylowman:
Closing for now - feel free to reopen

PR #292, opened 2022-11-02
Tensor5d6d

M1ngXU:
Impl's should be already impl'ed, just added (PAINFULLY ARGHHHH) all the broadcast/reduce impls. Added `tensor6d`-feature since this adds about 80% compilation time (18s -> 33s) due to all the permutation macro expansion (6! = 720 functions?).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,7 @@ libc = { version = "0.2", default-features = false, optional = true }
 default = ["std", "numpy"]
 std = ["no-std-compat/std", "rand/std", "rand_distr/std"]
 nightly = []
+tensor6d = []
 numpy = ["dep:zip", "std"]
 cblas = ["dep:cblas-sys", "dep:libc"]
 intel-mkl = ["cblas"]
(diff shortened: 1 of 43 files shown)


M1ngXU:
hopefully i did not cause any merging issues ....

M1ngXU:
is there actually the need to impl select?

M1ngXU:
ok sorry for opening a PR that quickly, i just found another block of impls missening

M1ngXU:
i *hope* that i did all the impls now

coreylowman:
Wow omg this is awesome!!

> Added tensor6d-feature since this adds about 80% compilation time (18s -> 33s)

Is this compile times for tests/examples? I'm curious if this would increase compile times for an external dependency

coreylowman:
**Really** appreciate the hard work with this - thanks to your investigation here the 5d6d permutations in the #304 are much more lightweight, and don't increase compile times as much! Without this PR I wouldn't have figured out how to make lightweight marker traits for all the shape stuff. Seriously thank you for this investigation!!

Issue #291, opened 2022-11-02
Add nightly tests to github workflows


coreylowman:
The command to run locally is `cargo +nightly test --tests`. There is probably a way to translate this into the cargo github workflow command, I vaguely remember it being documented

JYudelson1:
I checked https://github.com/actions-rs/meta, and it seems to just be another matrix in the .yml file to run on both stable and nightly. Made a PR for it at #294

PR #294, opened 2022-11-03
Adding nightly to cargo-test

JYudelson1:
As per #291


coreylowman:
Thanks for the contribution!

PR #298, opened 2022-11-08
Cb/stack

cBournhonesque:
Add a Stack trait, similar to the existing Broadcast/Reduce traits


coreylowman:
Will need to be refactored a ton after the devices rewrite - closing for now, thanks for the initial work on this!

PR #299, opened 2022-11-25
Tensor 5d 6d ops

### LINKED ISSUE #284: Add tensor op impls for Tensor5D and Tensor6D
5d and 6d tensors were added in
https://github.com/coreylowman/dfdx/issues/283#issuecomment-1300215357, but they don't have any sugar impls for the tensor ops. These are basically just macro invocations for all the tensor_ops

bbenshalom:
fixed #284 - implemented Tensor5D and Tensor6D broadcasting, reduction, permutation and additional tensor_ops. I added tests where I felt it wouldn't harm the readability of the entire file (for example, checking all possible permutations of Tensor6D is 720 lines of code... I assume it's undesired).

### DIFF
--- a/src/devices/broadcast_reduce/accumulator.rs
+++ b/src/devices/broadcast_reduce/accumulator.rs
@@ -126,3 +126,49 @@ pub(super) fn accum4d<A, L, R, const M: usize, const N: usize, const O: usize, c
         }
     }
 }
+
+#[inline(always)]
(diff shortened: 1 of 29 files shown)


coreylowman:
Closing this as most of this was merged into #304. A new PR will need to be added for 5d/6d reductions, which didn't make it in.

Appreciate this work!

PR #322, opened 2022-12-16
Adding Cuda device and skeleton cuda kernel impls

### LINKED ISSUE #321: Cuda device skeleton
Related to #9 

Todos here:
- [x] Add `Cuda` that is similar to `Cpu`. This will have StdRng as well as various structs from cudarc
- [x] Add `CudaArray` that is similar to `StridedArray` except it uses `Arc<CudaSlice<E>>` instead of `Arc<Vec<E>>`
- [x] Add feature flag `cuda` that adds dependency for cudarc
- [x] Add feature flag `test-cuda` that adds dependency for cudarc
- [x] `build_test_device!()` macro and `TestDevice` type should use `test-cuda` to create `Cuda` instead of `Cpu`
- [x] Add impl DeviceStorage for Cuda
- [x] impl tensor creation methods for Cuda
- [x] Add todo impls for all tensor_op kernels

Goal: be able to run `cargo test --tests --features test-cuda` and have it compile but fail at test time.

coreylowman:
Resolves #321


coreylowman:
> 

Nope, writing some initial kernels in the https://github.com/coreylowman/dfdx/tree/kernels branch first to make sure all of this actually works. Should be able to merge soon, still working on some improvements to cudarc base on kernel stuff

PR #331, opened 2023-01-06
Implementing abs/exp/div/sum_to cuda kernels

### LINKED ISSUE #184: CUDA kernels JIT vs compile time compilation
Should CUDA kernels be JIT compiled at runtime or somehow compiled when the program is built? Best case we can support both of these easily via a feature flag or something else. JIT would be nice for quicker builds, but pays cost at runtime.

coreylowman:
Also adds some necessary scaffolding for building/implementing them.

Resolves #184


src/tensor_ops/sum_to/sum_to.cu
```diff
@@ -0,0 +1,65 @@
...
+
+    if (i >= numel) {
+        return;
+    }
+
+    unsigned int inp_strided_i = get_strided_index(i, num_dims, dims, inp_strides);
+    auto tmp = inp[inp_strided_i];
+
+    unsigned int out_strided_i = get_strided_index(i, num_dims, dims, out_strides);
+    atomicAdd(out + out_strided_i, tmp);
```
coreylowman, line 37: While this is okay to start with, this will be really bad performance when reducing large tensors to 1 element. Need to add an issue to create a special kernel for that case (and there are standard ways to write high performant ones of those)

src/tensor_ops/sum_to/cuda_kernel.rs
```diff
@@ -1,27 +1,89 @@
...
+            self.dev
+                .load_ptx(PTX_SRC.into(), MODULE_NAME, &ALL_FN_NAMES)?;
+        }
+
+        let fwd_fn = self.dev.get_func(MODULE_NAME, FWD_FN_NAME).unwrap();
+
+        let dims: CudaSlice<usize> = self.dev.take_async(inp.shape.concrete().into())?;
+        let inp_strides: CudaSlice<usize> = self.dev.take_async(inp.strides.into())?;
+        let out_strides: Src::Concrete =
+            BroadcastStridesTo::<Src, Ax>::broadcast_strides(&dst, dst.strides());
```
coreylowman, line 35: the only difference between normal impl of forward and this is this call where the strides are broadcast. Maybe there is some way aroudn that?

src/tensor_ops/cuda_kernels.rs
```diff
@@ -0,0 +1,172 @@
...
+        if !self.dev.has_func(K::MODULE_NAME, K::FWD_FN_NAME) {
+            self.dev
+                .load_ptx(K::PTX_SRC.into(), K::MODULE_NAME, &K::ALL_FN_NAMES)?;
+        }
+
+        let numel = inp.data.len();
+        let mut storage = self.dev.alloc_zeros_async::<f32>(numel)?;
+
+        let fwd_fn = self.dev.get_func(K::MODULE_NAME, K::FWD_FN_NAME).unwrap();
+        let cfg = LaunchConfig::for_num_elems(numel as u32);
```
coreylowman, line 40: Should probably make a helper method for computing a good version of this launch config - need to take advantage of threads & blocks.

build.rs
```diff
@@ -8,10 +8,41 @@ fn main() {
...
+    cuda::build_ptx();
+
     #[cfg(feature = "intel-mkl")]
     intel_mkl::link().unwrap();
 }
 
+#[cfg(feature = "cuda")]
+mod cuda {
+    pub fn build_ptx() {
+        // TODO build ptx file in source tree and don't call nvcc if so
```
coreylowman, line 21: I think this can be done later, I'm not even sure if it's necessary. Once we have all the kernels in place we can see, but no need to complicate something that's pretty simple atm

src/tensor_ops/cuda_kernels.rs
```diff
@@ -0,0 +1,172 @@
...
+        grad_rhs: &mut Self::Storage<S, f32>,
+        grad_out: &Self::Storage<S, f32>,
+    ) -> Result<(), Self::Err> {
+        let bwd_fn = self.dev.get_func(K::MODULE_NAME, K::BWD_FN_NAME).unwrap();
+        let numel = lhs.shape.num_elements();
+
+        let dims: CudaSlice<usize> = self.dev.take_async(lhs.shape.concrete().into())?;
+        let lhs_strides: CudaSlice<usize> = self.dev.take_async(lhs.strides.into())?;
+        let rhs_strides: CudaSlice<usize> = self.dev.take_async(rhs.strides.into())?;
+        let out_strides: CudaSlice<usize> = self.dev.take_async(grad_out.strides.into())?;
```
coreylowman, line 153: These same values were also allocated in the forward call - a potential improvement for the future is pre-allocating them. Though these are only used in binary ops - if a tensor is only ever used in a unary op then it doesn't need to allocate these

Issue #333, opened 2023-01-07
Matmul/Conv2d/Pool2d cuda kernels


coreylowman:
These will be more challenging to implement, since they will rely on a lot of code and invoking cublas.

Also need to see if there's a way to simplify the implementation (potentially by having a device level matmul trait that matmul/conv2d can use)

coreylowman:
matmul tests blocked by #335

coreylowman:
This will require sub views of CudaSlices, which will be addressed in cudarc

M1ngXU:
all of them except matmul exist in cudnn

PR #342, opened 2023-01-07
matmul cuda kernels

coreylowman:
Progress on #333


src/tensor_ops/matmul/cuda_kernel.rs
```diff
@@ -1,13 +1,150 @@
...
+    n: N,
+    lhs: &A,
+    lhs_strides: [usize; 2],
+    rhs: &B,
+    rhs_strides: [usize; 2],
+    beta: f32,
+    out: &mut C,
+    out_strides: [usize; 2],
+) -> Result<(), CublasError> {
+    if out_strides[0] > 1 {
```
coreylowman, line 50: This function assumes one of the strides is 1, should add checks to ensure this.

src/tensor_ops/matmul/cuda_kernel.rs
```diff
@@ -1,13 +1,150 @@
...
+            transa,
+            transb,
+            m: n.size() as i32,
+            n: m.size() as i32,
+            k: k.size() as i32,
+            alpha: 1.0,
+            lda,
+            ldb,
+            beta,
+            ldc: n.size() as i32,
```
coreylowman, line 72: lda/ldb/ldc *should* be equal to the non-unit stride?

src/tensor_ops/matmul/cuda_kernel.rs
```diff
@@ -76,7 +453,66 @@ impl super::MatMatBrKernel<f32> for Cuda {
...
+            self.blas.gemm_strided_batched_async(
+                cfg,
+                rhs.data.as_ref(),
+                grad_out.data.as_ref(),
+                Arc::make_mut(&mut grad_lhs.data),
+            )?;
+        }
+        // TODO handle stride 0
+        let grad_rhs_buf = Arc::make_mut(&mut grad_rhs.data);
+        for b in 0..batch.size() {
```
coreylowman, line 491: add note that these have to be sequential since cublas sgemm doesn't use atomicAdd

PR #343, opened 2023-01-07
permute_to and broadcast_to cuda kernels

### LINKED ISSUE #335: Implement broadcast_to/permute_to cuda kernels
These both don't do anything on gpu for forward pass, and the backward pass is just summing the two arrays (without any striding)

coreylowman:
Resolves #335


coreylowman:
Current test failures:
```rust
---- tensor_ops::permute_to::tests::test_permute_4d_backwards stdout ----
thread 'tensor_ops::permute_to::tests::test_permute_4d_backwards' panicked at 'called `Result::unwrap()` on an `Err` value: Driver(DriverError(CUDA_ERROR_INVALID_VALUE, "invalid argument"))', src/tensor_ops/exp/mod.rs:33:24
```

coreylowman:
Above error was due to bug with LaunchConfig::for_num_elems, which was addressed in cudarc 0.5.1

Issue #341, opened 2023-01-07
Implement cuda kernels for unary ops


coreylowman:
These can mirror abs & exp which are already implemented:

- [x] abs
- [x] clamp
- [x] cos
- [x] exp
- [x] ln
- [x] nans_to
- [x] negate
- [x] powi
- [x] powf
- [x] relu
- [x] sigmoid
- [x] sin
- [x] square
- [x] tanh

ViliamVadocz:
Currently the test for `abs` fails on my machine. There is a small issue in the `abs_backward` implementation.

```
running 1 test
test tensor_ops::abs::tests::test_abs ... FAILED

failures:

---- tensor_ops::abs::tests::test_abs stdout ----
thread 'tensor_ops::abs::tests::test_abs' panicked at 'assertion failed: `(left == right)`
  left: `[0.2, 0.2, 0.0, -0.2, -0.2]`,
 right: `[-0.2, -0.2, 0.0, 0.2, 0.2]`', src\tensor_ops\abs\mod.rs:53:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I fixed it by changing line 27 in `abs.cu`. `signbit(x)` is true when the `x` is negative.
```diff
-    float dx = inp[i] == 0.0 ? 0.0 : (signbit(inp[i]) ? 1.0 : -1.0);
+    float dx = inp[i] == 0.0 ? 0.0 : (signbit(inp[i]) ? -1.0 : 1.0);
```

You could also write it a bit less confusingly like this:
```cpp
float dx = inp[i] == 0.0 ? 0.0 : copysignf(1.0, inp[i]);
```

ViliamVadocz:
If I understand correctly, the current kernels should be written for single-precision floating point numbers (`float` or `f32`). If so, it might be appropriate to use function specifically for floats such as `fabsf` or `expf` to avoid any accidental conversions to `double`s and back.

Here's a list of the available functions: [Single Precision Mathematical Functions](https://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH__SINGLE.html)

---

What are your thoughts on using [intrinsics](https://docs.nvidia.com/cuda/cuda-c-programming-guide/#intrinsic-functions)?

coreylowman:
Oooh good catch about the single precision stuff, I had no idea, so glad ya'll know 😀 What is the difference between intrinsics vs the non intrinsics?

coreylowman:
[...]

Issue #334, opened 2023-01-07
Implement cuda kernels for binary ops


coreylowman:
These can copy the setup for div cuda kernel, which is already implemented.

- [x] add
- [x] bce
- [x] div
- [x] huber_error
- [x] maximum
- [x] minimum
- [x] mul
- [x] sub

M1ngXU:
add, min, max, mul exist in cudnn

coreylowman:
Thanks @nkoppel

PR #346, opened 2023-01-10
Add cuda implementations for unary and binary tensor operations in #341 and #334

nkoppel:
This commit adds cuda kernels for the operations listed in #341 and #334. It also adds tests for huber_error and adds an "op" parameter to the binary kernels to allow huber_error to pass it's delta value to its kernel.

### DIFF
--- a/src/tensor_ops/exp/exp.cu
+++ b/src/tensor_ops/exp/exp.cu
@@ -10,7 +10,7 @@ extern "C" __global__ void exp_forward(
     if (i >= numel) {
         return;
     }
-    out[i] = exp(inp[i]);
+    out[i] = expf(inp[i]);
 }
 
 extern "C" __global__ void exp_backward(
@@ -24,5 +24,5 @@ extern "C" __global__ void exp_backward(
     if (i >= numel) {
         return;
     }
-    grad_inp[i] += exp(inp[i]) * grad_out[i];
+    grad_inp[i] += expf(inp[i]) * grad_out[i];
 }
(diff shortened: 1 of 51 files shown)


src/tensor_ops/exp/exp.cu
```diff
@@ -10,7 +10,7 @@ extern "C" __global__ void exp_forward(
     if (i >= numel) {
         return;
     }
-    out[i] = exp(inp[i]);
+    out[i] = expf(inp[i]);
```
coreylowman, line 13: Great catch here, thanks!

src/tensor_ops/huber_error/mod.rs
```diff
@@ -56,3 +56,34 @@ impl<S: Shape, E: Dtype, D: Device<E>, T: Tape<D>> Tensor<S, E, D, T> {
...
+
+#[cfg(test)]
+mod tests {
+    use crate::{
+        tensor::*,
+        tests::{assert_close, TestDevice},
+    };
+
+    #[test]
+    fn test_huber_error() {
```
coreylowman, line 68: Thanks for adding 😁

coreylowman (approved):
This is awesome!! Wow I was not expecting this haha what a super pleasant surprise to wake up to 😀 In another PR we can add `#[repr(C)]` to all the kernel op definitions

nkoppel:
Thank you for the quick response! One thing which may be of concern is that my powi.cu uses the "pow" function, which works on doubles rather than floats because the powi function doesn't seem to exist in cuda c and using doubles will prevent the exponent from being rounded.  If you don't end up changing this behavior, it should be documented, as `tensor.pow(4)` may be less performant than `tensor.pow(4.0)`.

It would probably be fine to use cuda c's powf function for this, as it shouldn't make a difference unless the exponent is in the millions.

coreylowman:
Yeah let's move to using powf and converting the exponent to a float, opening an issue for this in a sec

Issue #347, opened 2023-01-10
Add `#[repr(C)]` to all kernel op structs


coreylowman:
This is necessary to pass to cuda kernels

_Originally posted by @coreylowman in https://github.com/coreylowman/dfdx/pull/346#pullrequestreview-1242314646_

PR #354, opened 2023-01-10
Resolve #352 and #347

### LINKED ISSUE #352: Change powi cuda kernel to use powf
See thread by @nkoppel in https://github.com/coreylowman/dfdx/issues/346#issuecomment-1377480077

nkoppel:
Instead of having cuda c convert each exponent on the gpu, I thought it would be better to convert the exponent to a float on the cpu and call the powf kernel.


nkoppel:
The first commit resolves #352, and the second commit resolves #347.

coreylowman (approved):
Good changes 😁

Issue #349, opened 2023-01-10
Add tests for broadcasted/permuted binary ops


coreylowman:
Notably none of the unit tests for these test for different strides, but these are the most important places to test for this!

Can look something like:
```rust
    #[test]
    fn test_broadcasted_mul() {
        let dev: TestDevice = Default::default();
        let x = dev.tensor([1.0, 2.0, 3.0]);
        let y = dev.tensor([[-1.0, -0.5, -1.0 / 3.0], [2.0, 1.0, 0.0]]);
        let r = x.trace().broadcast() * y.clone();
        assert_eq!(r.array(), [[-1.0, -1.0, -1.0], [2.0, 2.0, 0.0]]);
        let g = r.exp().sum().backward();
        assert_close(&g.get(&x).array(), &[14.41023254, 7.20511627, -0.12262648]);
        assert_close(
            &g.get(&y).array(),
            &[
                [0.36787945, 0.73575890, 1.10363841],
                [7.38905621, 14.77811241, 3.0],
            ],
        );
    }
```

Except make the rhs permuted so we are covering both arguments with different strides

PR #355, opened 2023-01-11
Simplify cuda kernels for binary tensor operations.

nkoppel:
I realized that the math for calculating the stride indices in the binary operation cuda kernels was redundant, because the shape and strides of all input tensors are equivalent, and that the get_strided_index function ended up simply returning its input. I also changed the atomicAdds in the cuda kernels to normal adds, because each thread should modify exactly one value in grad_lhs and grad_rhs.

### DIFF
--- a/src/tensor_ops/add/binary_add.cu
+++ b/src/tensor_ops/add/binary_add.cu
@@ -1,71 +1,38 @@
 struct BinaryAddOp {};
 
-__device__ unsigned int get_strided_index(
-    unsigned int idx,
-    const size_t num_dims,
-    const size_t *dims,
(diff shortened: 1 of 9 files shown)


nkoppel:
All relevant tests in the mod.rs files are still passing when using test-cuda, so these changes should be sound.

nkoppel:
Never mind, this breaks using broadcasted arrays as arguments to binary operations.

coreylowman:
Yeah when doing broadcast I realized there's no unit tests that actually ensure binary ops work with broadcasted inputs. That's what #349 is for.

PR #357, opened 2023-01-12
Add tests for binary broadcasted add and fix bugs to allow them to pass.

nkoppel:
Implements tests for binary addition on broadcasted tensors and fixes two bugs which caused them to fail with gpu tensors. Works towards #349.

### DIFF
--- a/src/tensor/cuda/device.rs
+++ b/src/tensor/cuda/device.rs
@@ -115,7 +115,7 @@ impl DeviceStorage for Cuda {
         storage: &Self::Storage<S, E>,
     ) -> Result<Self::Storage<S, E>, Self::Err> {
         let numel = storage.shape.num_elements();
-        let strides: S::Concrete = storage.shape.strides();
+        let strides: S::Concrete = storage.strides;
         Ok(Self::Storage {
             data: Arc::new(self.dev.take_async(std::vec![Default::default(); numel])?),
             shape: storage.shape,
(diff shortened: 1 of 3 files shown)


src/tensor/cuda/device.rs
```diff
@@ -115,7 +115,7 @@ impl DeviceStorage for Cuda {
         storage: &Self::Storage<S, E>,
     ) -> Result<Self::Storage<S, E>, Self::Err> {
         let numel = storage.shape.num_elements();
-        let strides: S::Concrete = storage.shape.strides();
+        let strides: S::Concrete = storage.strides;
```
coreylowman, line 118: nice catch!

src/tensor_ops/cuda_kernels.rs
```diff
@@ -146,7 +146,7 @@ impl<K: BinaryOpCudaKernel + AsKernelParam> BinaryKernel<K, f32> for Cuda {
         grad_out: &Self::Storage<S, f32>,
     ) -> Result<(), Self::Err> {
         let bwd_fn = self.dev.get_func(K::MODULE_NAME, K::BWD_FN_NAME).unwrap();
-        let numel = lhs.shape.num_elements();
+        let numel = grad_out.shape.num_elements();
```
coreylowman, line 149: This shouldn't be necessary since they all have the same shape. What was happening before this was changed?

nkoppel: Yeah, my bad. I think I mistook that as being the number of elements physically stored in lhs, rather than the number of virtual elements. I'll revert that change.

PR #359, opened 2023-01-13
Add select and gather cuda kernels.

### LINKED ISSUE #338: Implement select_to/gather_to cuda kernels
These will be slightly more involved. Probably will be the most annoying cuda kernels to write.

nkoppel:
Adds cuda kernels for the SelectTo and GatherTo operations. Note that striding logic is not used to compute out_i in the gather kernel because the output tensor is never broadcasted, and because cudarc does not support calling functions with more than 12 arguments. Resolves #338.


coreylowman (approved):
nice work!!

PR #363, opened 2023-01-15
Retain as much broadcasting as possible for binary tensor operations

### LINKED ISSUE #351: Do sum_to need to use atomicAdd in backward?
This would be in the situation where the **input** tensor is already broadcasted, it could be writing to the same location multiple times:

```rust
x.trace().broadcast::<_, Axis<1>>().sum_to::<_, Axis<0>>();
```

Related to #349

nkoppel:
This pr changes binary tensor operations to keep elements as deduplicated as possible while being able to store the result correctly. This will result in much faster operations in some cases, especially when running on the cpu. This pr also resolves #351, as passing broadcasted tensors to sum_to in the broadcasted add tests exposed that the backward operation adds to the same location in grad_lhs and grad_rhs multiple times. 

One concern I have is that the merge_strides function may not be in the best location. The merge_strides function computes the strides needed to store the result of a binary operation given the strides of the input tensors.


src/tensor_ops/ops.rs
```diff
@@ -91,3 +91,23 @@ pub(crate) fn try_binary_op<
...
+    let mut prod = 1;
+    let dims = shape.concrete();
+
+    for i in (0..S::NUM_DIMS).rev() {
+        if strides1[i] > 0 || strides2[i] > 0 {
+            strides1[i] = prod;
+            prod *= dims[i];
+        }
+    }
+    (prod, strides1)
```
coreylowman, line 112: So this function is essentially saying "if both array's stride is 0 in this dimension, keep it at 0"? E.g. if I take two tensors with dim 0 broadcasted, then the resulting tensor will also have dim 0 broadcasted.

nkoppel: Yes, each stride of the output tensor is only broadcasted if that dimension is broadcasted for both of the input tensors to ensure that the output tensor can hold all unique resulting values.

coreylowman: Gotcha, that makes sense. I like this change, but it does make what the strides will be a little harder to reason about.

Let's defer until after cuda stuff is more stable. Additionally, it doesn't look like pytorch does this optimization, and I would like to investigate why they didn't (it seems to be a pretty straightforward one):
```python
>>> import torch
>>> a = torch.randn(2, 3)
>>> b = torch.randn(2, 3)
>>> c = a.broadcast_to(5, 2, 3) + b.broadcast_to(5, 2, 3)
>>> c.stride()
(6, 3, 1)
```

coreylowman:
> This pr also resolves https://github.com/coreylowman/dfdx/issues/351, as passing broadcasted tensors to sum_to in the broadcasted add tests exposed that the backward operation adds to the same location in grad_lhs and grad_rhs multiple times.

Would you mind splitting this off into a separate PR since this big one will be delayed?

nkoppel:
I can, but this pr should be ready to be merged. It passes all relevant tests now, including the one I added to test adding permuted arguments.

PR #367, opened 2023-01-16
Modify sum_to cuda kernel to not need atomic adds in backwards

### LINKED ISSUE #351: Do sum_to need to use atomicAdd in backward?

nkoppel:
Changes sum_to cuda kernel to not need atomic adds by spawning exactly one thread for each physical element in input or grad_input, and multiplying each added number by the number of threads which would have previously added that number. Resolves #351.

### DIFF
--- a/src/tensor_ops/sum_to/sum_to.cu
+++ b/src/tensor_ops/sum_to/sum_to.cu
@@ -13,28 +13,43 @@ __device__ unsigned int get_strided_index(
     return strided_i;
 }
 
+__device__ unsigned int get_unstrided_index(
+    const unsigned int strided_i,
+    const size_t num_dims,
+    const size_t *dims,
+    const size_t *strides
+) {
+    unsigned int idx = 0;
+    for (unsigned int d = 0; d < num_dims; d++) {
+        idx *= dims[d];
+        idx += strides[d] == 0 ? 0 : (strided_i / strides[d]) % dims[d];
+    }
+    return idx;
+}
+
 // Accepts pre-broadcasted strides for both input & output.
 // So both inp & out are expected to be broadcasted to the same size.
 extern "C" __global__ void sum_to_forward(
     const size_t numel,
     const size_t num_dims,
+    const float mul,
     const size_t *dims,
     const float *inp,
     const size_t *inp_strides,
     float *out,
     const size_t *out_strides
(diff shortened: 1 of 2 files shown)


src/tensor_ops/sum_to/sum_to.cu
```diff
@@ -13,53 +13,69 @@ __device__ unsigned int get_strided_index(
...
 
-    unsigned int inp_strided_i = get_strided_index(i, num_dims, dims, inp_strides);
-    auto tmp = inp[inp_strided_i];
+    auto tmp = inp[inp_i];
 
-    unsigned int out_strided_i = get_strided_index(i, num_dims, dims, out_strides);
-    atomicAdd(out + out_strided_i, tmp);
+    unsigned int i = get_unstrided_index(inp_i, num_dims, dims, inp_strides);
+    unsigned int out_i = get_strided_index(i, num_dims, dims, out_strides);
+    atomicAdd(out + out_i, tmp * mul);
```
coreylowman, line 52: What does the `tmp * mul` do here? I see that mul is `let mul = (numel / inp_numel) as f32;`, but won't that scale all the values by that ratio?

nkoppel: Yes, this is valid because "mul" is the number of virtual elements represented by each physical element, and this change only calls a thread for each physical element of the tensor. I'll try to make this code easier to understand.

coreylowman: Nice readability improvements, much clearer!

Ohhh I see, so this works for backwards because we assume that all the output gradients are the same, meaning we can access a single one and multiply by the number of them?

I don't see why this works for forward pass though since we can't make the same assumption that all threads would be adding the same exact element. What am I missing?

coreylowman (approved):
nice changes! 🚀

src/tensor_ops/sum_to/sum_to.cu
coreylowman: Gotcha, thanks!

PR #368, opened 2023-01-16
Adding dynamic example.

### LINKED ISSUE #308: Add dynamic batch example for tensor ops & nn
Show case the new dynamic shapes!!

Narsil:
Fixes #308

Is that example good enough ?

@coreylowman

### DIFF
--- a/examples/01-tensor.rs
+++ b/examples/01-tensor.rs
@@ -1,7 +1,7 @@
 //! Intro to dfdx::arrays and dfdx::tensor
 
 use dfdx::{
-    shapes::{Rank1, Rank2, Rank3},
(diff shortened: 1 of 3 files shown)


Narsil:
> Yeah this is a great add! I think it'd be good to also something like this to the 03-nn.rs example after this line https://github.com/coreylowman/dfdx/blob/main/examples/03-nn.rs#L30

Added ! I took the liberty of adding it in the main doc too. Tell me if there's anything more to add.

coreylowman:
Perfect! Thanks for the contribution!

coreylowman (approved):
🚀

PR #380, opened 2023-01-19
Fix max_to and min_to edge case with negative zero

### LINKED ISSUE #379: Max_to cuda kernel bug with negative zero
Code added in #370 contains a bug to do with negative zero.

Test case:

```rust
#[test]
fn test_max_negative_zero() {
    let dev: TestDevice = Default::default();
    let t = dev.tensor([[-0.0, 0.0], [0.0, -0.0], [-0.0, -1.0], [0.0, -1.0]]);
    let r = t.trace().max::<_, Axis<1>>();
    assert_eq!(r.array(), [0.0, 0.0, -0.0, 0.0]);
    let g = r.sum().backward();
    assert_eq!(
        g.get(&t).array(),
        [[1.0, 1.0], [1.0, 1.0], [1.0, 0.0], [1.0, 0.0]]
    );
}
```

This test passes on cpu, but with the `test-cuda` feature it gives this failure:

```rust
---- tensor_ops::max_to::tests::test_max_negative_zero stdout ----
thread 'tensor_ops::max_to::tests::test_max_negative_zero' panicked at 'assertion failed: `(left == right)`
  left: `[0.0, 0.0, -1.0, 0.0]`,
 right: `[0.0, 0.0, -0.0, 0.0]`', src\tensor_ops\max_to\mod.rs:144:9
 ```
 
 This bug was mentioned in the StackOverflow page from which the code for the kernel was taken, along with a suggested fix: https://stackoverflow.com/a/72461459/10514840

ViliamVadocz:
Closes #379 

The fix was adapted from the StackOverflow answer mentioned in the issue.
I also added two tests that failed before and now succeed.

### DIFF
--- a/src/tensor_ops/max_to/max_to.cu
+++ b/src/tensor_ops/max_to/max_to.cu
@@ -1,10 +1,10 @@
 // atomicMax is not implemented for floats,
 // solution copied https://stackoverflow.com/questions/17399119/how-do-i-use-atomicmax-on-floating-point-values-in-cuda
 __device__ __forceinline__ float atomicMaxf(float * addr, float value) {
-    if (value >= 0) {
-        return  __int_as_float(atomicMax((int *)addr, __float_as_int(value)));
+    if (signbit(value)) {
+        return __uint_as_float(atomicMin((unsigned int *)addr, __float_as_uint(value)));        
     } else {
-        return __uint_as_float(atomicMin((unsigned int *)addr, __float_as_uint(value)));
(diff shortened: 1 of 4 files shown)


ViliamVadocz:
Do I need to do anything to make the coverage pass?

coreylowman:
> Do I need to do anything to make the coverage pass?

Nah sometimes it just fails, I've been thinking about removing it, not sure how much coverage matters anyway

coreylowman (approved):
Nice fixes and thanks for the new tests! 🚀

PR #381, opened 2023-01-20
Safetensors support.

Narsil:
I'm using this to do experiments feel free to close or ignore.
Main goal is to be able to use weights contained on https://hf.co
which doesn't include NPZ format.

All models supporting `safetensors`: https://huggingface.co/models?library=safetensors&sort=downloads


coreylowman:
Great idea! This'll be great for getting pretrained weights since I had no idea how to load them from pickle files

Narsil:
Nice, I'll open this PR up then. 

Biggest caveats I've seen:

There's a copy on write because safetensors expects a flat `[u8]` buffer. There seems to be a copy here: https://github.com/coreylowman/dfdx/pull/381/files#diff-a0868f528cbb396ebf73e5726654153f4ba44c8820003f936d809e8ae45e6cdaR35

It's usually not that bad on save, since it's not the biggest drawback, but still something to take into account when training large models.

In general I think we could aim for zero-copy loads with memory mapping on aligned files: https://github.com/coreylowman/dfdx/pull/381/files#diff-a0868f528cbb396ebf73e5726654153f4ba44c8820003f936d809e8ae45e6cdaR124
But not worth doing atm I think (That would mean having a borrowed buffer, which kills possibility of doing gradients stuff, just mentionning in case)

coreylowman:
Yeah I think given the tensors may be stored on other devices, getting a `&[u8]` slice seems unlikely. E.g. for cuda we're going to have to copy back to CPU `Vec<E>` first anyway.

src/tensor/safetensors.rs
```diff
@@ -0,0 +1,138 @@
...
+    tensors: BTreeMap<String, TensorData>,
+}
+
+impl Writer {
+    pub fn new() -> Self {
+        let tensors = BTreeMap::new();
+        Self { tensors }
+    }
+
+    pub fn add<S, T, E: Dtype + SafeDtype>(mut self, key: String, tensor: Tensor<S, E>) -> Self
```
coreylowman, line 27: Possible to add DeviceStorage generic to Tensor here? So we can add tensors from different devices 😄

Narsil: I did what I could, but it felt like type monkey coding. I added stuff until it stopped complaining. Not sure it's the best possible type :)

Narsil:
> Yeah I think given the tensors may be stored on other devices, getting a &[u8] slice seems unlikely. E.g. for cuda we're going to have to copy back to CPU Vec<E> first anyway.

Yes, and training is most likely going to occur on GPU most of the time. Still extra copies in general are usually becoming bottleneck really fast so I'm acutely aware of them.

Loading is happening much more than saving in practice (just like inference >> training overall) so as long as those are reduced to a minimum it's Ok.

Narsil:
Any new thoughts on this PR ?

src/tensor/safetensors.rs
```diff
@@ -0,0 +1,143 @@
...
+    tensors: BTreeMap<String, TensorData>,
+}
+
+impl Writer {
+    pub fn new() -> Self {
+        let tensors = BTreeMap::new();
+        Self { tensors }
+    }
+
+    pub fn add<S, T, E: Dtype + SafeDtype, D: DeviceStorage + CopySlice<E>>(
```
coreylowman, line 28: Is CopySlice needed for this? Might be able to just use DeviceStorage

src/tensor/safetensors.rs
```diff
@@ -0,0 +1,143 @@
...
+        mut self,
+        key: String,
+        tensor: Tensor<S, E, D, T>,
+    ) -> Self
+    where
+        S: Shape,
+        <D as DeviceStorage>::Storage<S, E>: HasUnitType<Unit = E> + AsVec,
```
coreylowman, line 35: As of #465, you shouldn't need this bound anymore!

src/tensor/safetensors.rs
```diff
@@ -0,0 +1,143 @@
...
+
+    pub fn load<'a>(
+        &mut self,
+        tensors: &SafeTensors<'a>,
+        key: &str,
+    ) -> Result<(), SafeTensorError> {
+        let tensor = tensors.tensor(key)?;
+        let v = tensor.data();
+        let num_bytes = std::mem::size_of::<E>();
+        if (v.as_ptr() as usize) % num_bytes == 0 {
```
coreylowman, line 126: How sure are you that this is correct? Is this expected use for safetensors? I see std::ptr has is_aligned_to but its on nightly unfortunately.

coreylowman (commented):
Looks good to me. After #460 it will be straightforward to impl this for nn layer as well

coreylowman (commented):
TensorCollection api was just merged, which greatly simplified nn::SaveToNpz/LoadFromNpz. Should be really easy to add a safetensors loader/saver for nn layer now as well!

src/tensor/safetensors.rs
coreylowman: Gotcha, makes sense, we can keep for now and then move to using is_aligned_to whenever it becomes stable

Narsil:
Great addition ! 

I updated everything ! We can now simply load/save entire modules. Nice !

Cargo.toml
```diff
@@ -33,13 +33,17 @@ cblas-sys = { version = "0.1.4", default-features = false, optional = true }
 libc = { version = "0.2", default-features = false, optional = true }
 cudarc = { version = "0.7.2", default-features = false, optional = true }
 num-traits = { version = "0.2.15", default-features = false }
+# safetensors = { version = "0.2", default-features = false, optional = true }
+safetensors = { git = "https://github.com/huggingface/safetensors", default-features = false, optional = true }
```
coreylowman, line 37: Was this a holdover from developing? Which should be used?

coreylowman (changes requested):
🔥 will be good to merge after version fix in Cargo.toml

Cargo.toml
Narsil: I need to release `0.3` . Should happen this week. (I will release python release candidate first, there has been a relatively big change in how things are saved, everythng backward compatible, but files on disk will be different now, to get alignement properly done).

Narsil:
@coreylowman I fixed with the released version.

coreylowman:
@Narsil looks good. didn't this used to have an impl for tensor collections?

Narsil:
I'm as confused as you, I simply rebased. Doing the work again since I can't find that work anymore. I'm guessing I screwed my rebase but I can't figure out where the original code lives.

Narsil:
I added the test with the feature as I'm seeing a super weird trait issue in the Booleans module, which doesn't seem linked to my changes ? Any ideas: https://github.com/coreylowman/dfdx/actions/runs/4364586004/jobs/7632084283

examples/safetensors-save-load.rs
```diff
@@ -0,0 +1,44 @@
...
+
+    let mut gpt2 = dev.build_module::<Linear<728, 50257>, f32>();
+    let filename = "gpt2.safetensors";
+    let f = std::fs::File::open(filename).expect("Couldn't read file, have you downloaded gpt2 ? `wget -O gpt2.safetensors https://huggingface.co/gpt2/resolve/main/model.safetensors`");
+    let buffer = unsafe { MmapOptions::new().map(&f).expect("Could not mmap") };
+    let tensors = SafeTensors::deserialize(&buffer).expect("Couldn't read safetensors file");
+
+    gpt2.weight
+        .load_safetensor(&tensors, "wte.weight")
+        .expect("Could not load tensor");
```
coreylowman, line 38: Awesome example

coreylowman (approved):
LGTM! Don't know why the boolean test is failing, its passing locally for me. Will resolve separately. Thanks for the contribution!

PR #382, opened 2023-01-20
Efficient cuda kernels for reductions

### LINKED ISSUE #332: Optimize cuda sum_to kernel for full reductions
Currently the sum_to reduction (& in the future max/min/etc) use atomicAdd to handle concurrent writes to the same output variable.

For reductions that reduce to 1 number this will result in a lot of contention for writes.

There should be a special kernel for reductions to 1 value based on standard reduction kernels. The sum_to kernel implementation can check at runtime if the size of the shape is 1 and execute the different kernel

_Originally posted by @coreylowman in https://github.com/coreylowman/dfdx/pull/331#discussion_r1063738025_

nkoppel:
So far, implements a more efficient sum_to kernel that will have a maximum write contention of the number of blocks (groups of 1024 threads) running concurrently. Operations within each block scale with `log2(min(chunk_size, block_size))`. Resolves #332, and will depend on #380 for @ViliamVadocz 's fix to atomicMaxf and atomicMinf.


nkoppel:
I've now finished the min_to and max_to kernels.

src/tensor_ops/max_to/cuda_kernel.rs
```diff
@@ -80,11 +82,15 @@ impl super::MaxReduceKernel<f32> for Cuda {
             BroadcastStridesTo::<Src, Ax>::broadcast_strides(&grad_out.shape, grad_out.strides);
         let out_strides: CudaSlice<usize> = self.dev.take_async(out_strides.into())?;
 
-        let inp_numel = grad_inp.shape.num_elements();
-        let cfg = LaunchConfig::for_num_elems(inp_numel as u32);
+        let physical_numel = grad_inp.data.len();
+        let virtual_numel = grad_inp.shape.num_elements();
+        let elems_per_thread = (virtual_numel / physical_numel) as f32;
+
+        let cfg = LaunchConfig::for_num_elems(physical_numel as u32);
```
coreylowman, line 89: Do we need to configure shared memory here? There's this field for LaunchConfig that hasn't been used yet anywhere https://docs.rs/cudarc/latest/cudarc/driver/safe/struct.LaunchConfig.html#structfield.shared_mem_bytes

src/tensor_ops/internal_reshapes.rs
```diff
@@ -0,0 +1,31 @@
+#[cfg(feature = "cuda")]
+use crate::prelude::Axes;
+#[cfg(feature = "cuda")]
+use std::vec::Vec;
+
+/// Moves all axes in Ax to the end of dims and strides and removes broadcasted dimensions
+/// so that a cuda kernel called for each physical element of the input tensor will place elements
+/// to be reduced with each other next to each other in memory.
```
coreylowman, line 8: > will place elements to be reduced with each other next to each other in memory.

How exactly does this happen? Changing strides won't actually move stuff around in memory, so I guess i'm not seeing how we can depend on the sequential addressing used for the chunk_min/chunk_max/chunk_sum.

What am I missing?

src/tensor_ops/max_to/max_to.cu
```diff
@@ -31,34 +45,88 @@ extern "C" __global__ void fill_with(float *buf, float value, const size_t numel
...
+    const size_t numel,
+    const size_t chunk_len,
+    const float data,
+    float* out
+) {
+    __shared__ float buf[1024];
+    // assumes that threads where i >= numel have already exited
+    unsigned int i = blockIdx.x * blockDim.x + threadIdx.x;
+    unsigned int block_i = threadIdx.x;
+    buf[block_i] = data;
```
coreylowman, line 71: Ahhh is this where the strides come into play? The `data` value here is from inp[get_strided_index(...)], so each thread storing data in this shared buffer will be controlled by the strides.

coreylowman (commented):
Looks great, thanks for this contribution! Just have some questions to make sure I'm following 🚀

src/tensor_ops/max_to/cuda_kernel.rs
nkoppel: Per [nvidia's documentation](https://developer.download.nvidia.com/compute/DevZone/docs/html/C/doc/html/group__CUDA__EXEC_gb8f3dc3031b40da29d5f9a7139e52e15.html), this parameter controls the amount of dynamic shared memory available to each thread block. However, we currently don't have to configure dynamic shared memory because we use statically sized shared memory, and we only need 1024 elements at maximum because this is a limit set by CUDA. Despite this code working, it relies on LaunchConfig's [for_num_elements](https://docs.rs/cudarc/latest/src/cudarc/driver/safe.rs.html#639-647) setting a block size of 1024 to not waste memory.

src/tensor_ops/max_to/max_to.cu
nkoppel: Yes, `data` is an array of values passed in to `chunk_sum` and others through each thread's registers.

coreylowman: Gotcha, very nice!

coreylowman (approved):
Awesome changes, thanks for contribution!

Issue #377, opened 2023-01-18
impl `Shape` for `[usize; N]`


coreylowman:
It would be nice to use usize arrays for tensors with entirely dynamic shapes:

```rust
let dev: Cpu = Default::default();
let a = dev.zeros_like(&[2, 4, 6]);
```

nkoppel:
This should be closed by the merge of #385.

PR #385, opened 2023-01-21
Implement Shape for arrays (#377)

nkoppel:
This pr implements the Shape trait for arrays of length 1-6, resolving #377. It also reformats many of the macro definitions involved with macro definitions to make them, in my opinion, more readable. It doesn't implement the HasSameNumelAs trait for arrays, as this is to my knowledge impossible. Perhaps we could implement runtime-checked versions of reshape_to and flatten, which would also have the benefit of working on stable.

### DIFF
--- a/src/shapes/axes.rs
+++ b/src/shapes/axes.rs
@@ -85,13 +85,20 @@ pub trait HasAxes<Ax> {
 }
 
 macro_rules! impl_has_axis {
-    (($($Vars:tt),*), $Axis:tt) => {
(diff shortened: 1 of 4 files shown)


coreylowman:
Agree about runtime-checked versions of reshape_to/flatten - perhaps do something like `checked_reshape() -> Option<V>`? Similar to things from std lib that have checked version

coreylowman (approved):
Nice changes!

Issue #306, opened 2022-12-11
Add comparison tensor_ops


coreylowman:
- [ ] eq
- [ ] ne
- [ ] gt
- [ ] gte
- [ ] lt
- [ ] lte

functions should look like
```rust
fn eq<S: Shape, E: Dtype, D: Device<E>, T>(a: Tensor<S, E, D, T>, b: Tensor<S, E, D, T>) -> Tensor<S, bool, D, T>
```

coreylowman:
Note that you won't be able to backprop through these - so they will always return a tensor with a NoneTape. This is the same as pytorch behavior here

coreylowman:
There should also be a way for comparing against scalar values, but that can come later

ViliamVadocz:
I assume this will also require `Device` with `Unit`, not `Dtype`. 

Currently code like this does not compile because of the trait bound.
```rust
fn eq<S: Shape, E: Unit, D: Device<E>, T>(
    a: Tensor<S, E, D, T>,
    b: Tensor<S, E, D, T>,
) -> Tensor<S, bool, D, T> {
  todo!()
}
```
```
8 | fn eq<S: Shape, E: Unit, D: Device<E>, T>(
  |                             ^^^^^^^^^ the trait `shape::Dtype` is not implemented for `E`
  ```

coreylowman:
Ah yeah my mistake in the description. You'll need to use DeviceStorage instead of Device. (Device is just DeviceStorage + all the traits for tensor ops, but will require `E: Dtype`). We can't remove that restriction because pretty much every other tensor op requires Dtype for gradients.

So the idea would be generally:

```rust
// step 1: add an EqKernel trait
// NOTE: E here is Unit instead of Dtype
trait EqKernel<E: Unit>: DeviceStorage {
     fn forward(...)
     // NOTE: won't need a backward method since these won't support backprop
}

// step 2: add eq method to Tensor
impl<S: Shape, E: Unit, D: EqKernel<E>, T: Tape<D>> Tensor<S, E, D, T> {
    pub fn eq(self, other: Self) -> Tensor<S, bool, D, NoneTape> {
         todo!("call kernel forward methods on device");
    }
}

// step 3: add eq function that calls tensor method. This is mainly for discovery for functions & docstrings
[...]

PR #386, opened 2023-01-21
Add comparison tensor operations

ViliamVadocz:
Works towards #306.
Depends on https://github.com/coreylowman/cudarc/pull/60.

I am creating this as a draft to get feedback on the structure before implementing other operations (although this should be quite straightforward now).

Let me know what you think.
I will add documentation once implementation is finalized.


ViliamVadocz:
Some things to consider and choices I would like feedback on:
- Should `Unit` require `PartialEq`? At the moment I enforce that at the implementation for `EqKernelOp` as an additional requirement.
- I used a generic to specify the type of operation. I know that other kernel operations (such as `UnaryKernel` and `BinaryKernel`) pass in an empty struct. Should this also be done for comparison operations?
- I placed the comparison operations in a sub-directory `cmp` I did this so that I could separate the comparison-specific kernel support code from the code for other operations which require `Dtype`. I think this is a good solution but perhaps a different structure is preferred?

nkoppel:
>  I used a generic to specify the type of operation. I know that other kernel operations (such as `UnaryKernel` and `BinaryKernel`) pass in an empty struct. Should this also be done for comparison operations?
It should be fine not to. I added these struct parameters to the binary operations in #346 only because of huber_error's delta parameter, and comparison operations should not have this kind of parameters.

You should probably implement unary versions of these operations, because this is consistent with the rest of the api and will probably be the more common use case.

While tensors of booleans are currently useless with the rest of the api, I think it is best to keep the output as tensors of booleans and to create operations for them. These could include inversion (`!`), and (`&`), or (`|`), xor (`^`), and masking operations for use with other tensors.

ViliamVadocz:
> You should probably implement unary versions of these operations, because this is consistent with the rest of the api and will probably be the more common use case.

I don't understand how comparison could be unary. Did you mean something like comparing to a scalar? Or perhaps you meant that the comparison functions should be methods on the tensor type? (They are.)

nkoppel:
Sorry for the confusion, I meant that you should add operations that act like ScalarAdd, comparing each value within the tensor with a single number, like `<tensor>.eq(0.0)`.

coreylowman (commented):
Looks like a great start! I like the addition of CmpKernel/CmpOpCpuKernel/CmpOpCudaKernel for reducing code. Since you added the above and the actual kernel implementations look pretty small, I think combining them into the same file so the structure is the same as the other ops should be good:

```
- tensor_ops/
    - cmp/
        - mod.rs
        - cuda_kernels.rs
        - cpu_kernels.rs
        - cmp.cu
```

where they would contain the following:
1. mod.rs - all the tensor methods/functions & tests
2. cpu_kernels.rs
    1. The trait definition for CmpOpCpuKernel
    2. The trait impl for `impl<Op: CmpOpCpuKernel<E>, E: Unit> CmpKernel<Op, E> for Cpu {`
    3. all the `impl CmpOpCpuKernel<E> for EqKernelOp` for all the ops (Eq/Lt, etc)
3. cuda_kernels.rs - similar to cpu kernels but for the cuda stuff
4. cmp.cu all the forward functions for each op

src/tensor_ops/cmp/eq/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+        self.try_eq(other).unwrap()
+    }
+}
+
+pub fn eq<S: Shape, E: Unit, D: CmpKernel<EqKernelOp, E>, T: Tape<D>>(
+    lhs: Tensor<S, E, D, T>,
+    rhs: Tensor<S, E, D, T>,
+) -> Tensor<S, bool, D, NoneTape> {
+    lhs.eq(rhs)
+}
```
coreylowman, line 30: Thoughts on making all the tensors reference instead of value? It wouldn't be consistent with other apis, but cmp isn't consistent in the first place since you can't backprop through it. Perhaps taking by reference would make that more clear (and reduce cloning where we can)?

```suggestion
impl<S: Shape, E: Unit, D: CmpKernel<EqKernelOp, E>, T: Tape<D>> Tensor<S, E, D, T> {
    pub fn try_eq(&self, other: &Self) -> Result<Tensor<S, bool, D, NoneTape>, D::Err> {
        try_cmp_op(self, other)
    }

    pub fn eq(&self, other: &Self) -> Tensor<S, bool, D, NoneTape> {
        self.try_eq(other).unwrap()
    }
}

pub fn eq<S: Shape, E: Unit, D: CmpKernel<EqKernelOp, E>, T: Tape<D>>(
    lhs: &Tensor<S, E, D, T>,
    rhs: &Tensor<S, E, D, T>,
) -> Tensor<S, bool, D, NoneTape> {
    lhs.eq(rhs)
}
```

src/shapes/shape.rs
```diff
@@ -8,6 +8,17 @@ pub trait Unit:
...
+impl Unit for u16 {}
+impl Unit for u32 {}
+impl Unit for u64 {}
+impl Unit for u128 {}
+impl Unit for isize {}
+impl Unit for i8 {}
+impl Unit for i16 {}
+impl Unit for i32 {}
+impl Unit for i64 {}
+impl Unit for i128 {}
```
coreylowman, line 21: Woohoo!

ViliamVadocz:
I added the rest of the comparison operations as well as scalar operations. I was unsure about what to name the scalar comparisons, but I went with `scalar_op` where `op` is replaced by one of `eq`, `ne`, `ge`, etc.

I was wondering whether I should also implement `PartialEq` and `PartialCmp` for `Tensor` so that it would be possible to use the actual operators like `<`, etc.

nkoppel:
The `<`, `==`, and other comparison operations can, unfortunately, only output single booleans. However, I think it's a good idea to implement PartialEq and PartialOrd for tensors. The cuda implementation for PartialEq should boil down to a reduction, but the kernel for ParitalOrd will need to prioritize earlier elements in the tensors, which will require something like a custom atomic operation.

src/tensor_ops/cmp/mod.rs
```diff
@@ -0,0 +1,458 @@
...
+        &self,
+        lhs: &Self::Storage<S, E>,
+        rhs: &Self::Storage<S, E>,
+    ) -> Result<Self::Storage<S, bool>, Self::Err>;
+}
+
+fn try_cmp_op<Op, S: Shape, E: Unit, D: CmpKernel<Op, E>, T: Tape<D>>(
+    lhs: &Tensor<S, E, D, T>,
+    rhs: &Tensor<S, E, D, T>,
+) -> Result<Tensor<S, bool, D, NoneTape>, D::Err> {
```
coreylowman, line 22: Will need to add a runtime shape check as I discovered in #420 
```suggestion
) -> Result<Tensor<S, bool, D, NoneTape>, D::Err> {
    assert_eq!(lhs.shape(), rhs.shape());
```

src/tensor_ops/cmp/mod.rs
```diff
@@ -0,0 +1,458 @@
...
+    fn test_eq() {
+        test_cmp(
+            [[1.0, 2.0, 3.0], [4.0, 5.0, 0.0]],
+            [[0.0, 2.0, -3.0], [4.0, 0.5, -0.0]],
+            |a, b| a.eq(b).array(),
+            [[false, true, false], [true, false, true]],
+        );
+    }
+
+    #[cfg(not(feature = "cuda"))]
```
coreylowman, line 289: Ahhh this is necessary because cuda kernels are only doing f32 dtype right now? Can you add a `// TODO` comment to remove the feature once they are added?

coreylowman (changes requested):
This looks great! Structure is perfect, and a lot of good stuff in here. Only a couple things to add and then its good to go 🚀

src/tensor_ops/cmp/mod.rs
ViliamVadocz: No it's needed because Cuda does not support integers. These tests are making sure that the operations behave as expected for types which are `Unit` but not `Dtype`.

ViliamVadocz:
Please note that the CUDA code relies on `bool` implementing `ValidAsZeroBit`. This was already merged in `cudarc` (https://github.com/coreylowman/cudarc/pull/60), but a new version of the crate has not been published yet.

If you try compiling this branch with `test-cuda` you will get errors unless you also change the `cudarc` dependency to use the main branch from git. Probably best to wait with merging this until a new version of `cudarc` is published.

src/tensor_ops/cmp/mod.rs
```diff
@@ -456,4 +456,13 @@ mod tests {
             [[true, true], [false, true]],
         );
     }
+
+    #[test]
+    #[should_panic]
```
coreylowman, line 467: Nice test 😀

coreylowman (approved):
Will merge once cudarc releases support for bools

coreylowman:
Thanks for the contribution!

Issue #390, opened 2023-01-23
Slow compilation of VGG-16


leofidus:
Using the current dfdx master on the current nightly compiler, the code below takes 15 minutes to compile in release mode (after compiling all dependencies). Commenting out either the block labeled `conv block 5` or `head` drops the compile time to one minute.

```rust
#![feature(generic_const_exprs)]
use dfdx::{prelude::*, tensor::Tensor};

/// VGG-16
type Mlp = (
    // conv block 1
    (
        (Conv2D<3, 64, 3, 1, 1>, ReLU),
        (Conv2D<64, 64, 3, 1, 1>, ReLU),
        MaxPool2D<2, 2>,
    ),
    // conv block 2
    (
        (Conv2D<64, 128, 3, 1, 1>, ReLU),
        (Conv2D<128, 128, 3, 1, 1>, ReLU),
        MaxPool2D<2, 2>,
    ),
    // conv block 3
    (
        (Conv2D<128, 256, 3, 1, 1>, ReLU),
        (Conv2D<256, 256, 3, 1, 1>, ReLU),
        (Conv2D<256, 256, 3, 1, 1>, ReLU),
        MaxPool2D<2, 2>,
    ),
    // conv block 4
    (
        (Conv2D<256, 512, 3, 1, 1>, ReLU),
        (Conv2D<512, 512, 3, 1, 1>, ReLU),
        (Conv2D<512, 512, 3, 1, 1>, ReLU),
        MaxPool2D<2, 2>,
    ),
    // conv block 5
    (
        (Conv2D<512, 512, 3, 1, 1>, ReLU),
        (Conv2D<512, 512, 3, 1, 1>, ReLU),
        (Conv2D<512, 512, 3, 1, 1>, ReLU),
        MaxPool2D<2, 2>,
    ),
    // head
    (
        (Conv2D<512, 4096, 7, 1, 0>, ReLU),
        (Conv2D<4096, 4096, 1, 1, 0>, ReLU),
        (Conv2D<4096, 2622, 1, 1, 0>, Flatten2D, Softmax),
    ),
);

fn main() {
    let dev: Cpu = Default::default();
    let mlp: Mlp = dev.build_module();
    let x: Tensor<Rank3<3, 224, 224>> = dev.zeros();
    let y = Module::forward(&mlp, x);
}
```

I tried deeper tuples (grouping block 1-3 and 4-head), but that didn't seem to have an appreciable effect.

coreylowman:
[...]

PR #395, opened 2023-01-24
Add VGG16 example

leofidus:
A slightly polished version of my code in #390. The changes are mostly in making it cross-platform, and making it more obvious how to use the network (putting data in and getting it out).

Starting another thread isn't very elegant, but Windows has a 1MB stack for the main thread and I also didn't want to just do `dev.zeros()` and completely leave data loading as an exercise to the reader. Is there a better way to initialize large tensors?

### DIFF
--- a/examples/nightly-vgg16.rs
+++ b/examples/nightly-vgg16.rs
@@ -1,0 +1,89 @@
+//! This example implements VGG16, a deep convolutional network popular in facial recognition
+#![cfg_attr(feature = "nightly", feature(generic_const_exprs))]
+
+#[cfg(feature = "nightly")]
+use dfdx::prelude::*;
+
+#[cfg(feature = "nightly")]
+type Vgg = (
+    // conv block 1
+    (
+        (Conv2D<3, 64, 3, 1, 1>, ReLU),
+        (Conv2D<64, 64, 3, 1, 1>, ReLU),
+        MaxPool2D<2, 2>,
+    ),
+    // conv block 2
+    (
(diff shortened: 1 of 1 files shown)


examples/nightly-vgg16.rs
```diff
@@ -0,0 +1,89 @@
...
+    // setup, takes about 5 seconds
+    let dev: Cpu = Default::default();
+    let vgg: Vgg = dev.build_module();
+
+    // load weights, if you have them
+    // vgg.load("vgg_weights.npz").unwrap()
+
+    // this should be a 224x224 RGB image, but for demonstration let's just use zeros
+    let image = [[[0.0f32; 224]; 224]; 3];
+    let x: Tensor<Rank3<3, 224, 224>> = dev.tensor(image);
```
coreylowman, line 73: I think using `dev.sample_uniform()` or `dev.sample_normal()` if you don't want to use zeros. Otherwise just do `dev.zeros()`. If any of those overflow the stack should open a ticket!

This example's main purpose is to just show how to construct the vgg model type, so it shouldn't necessarily worry about how to load data properly or things like that.

leofidus: Well, the library user has to load data at some point to use the model, and if they are anything like me they skipped the 01-tensors example and are left wondering how to do that :D

Imho one of the strengths of examples over other documentation is showing how to assemble things end-to-end, instead of just the exciting part in the middle

coreylowman: Haha fair enough - what do you think would be a compelling example for loading data for this example?

PR #397, opened 2023-01-24
Adding GeLU operator (used in Gpt2)

Narsil:
- Added the operations as found here:

https://github.com/pytorch/pytorch/blob/acdd462b1a070790799ce4623ce8ecc83e197e81/torch/_decomp/decompositions.py#L217
https://github.com/pytorch/pytorch/blob/acdd462b1a070790799ce4623ce8ecc83e197e81/caffe2/operators/gelu_op.cu#L84
https://pytorch.org/docs/stable/generated/torch.nn.GELU.html

- Added both cuda and cpu versions.
- Added a `simplify` helper method, because the results are slightly
  different between cuda and cpu (this is relatively common IMO).


src/tensor_ops/gelu/cpu_kernel.rs
```diff
@@ -0,0 +1,33 @@
+use crate::tensor_ops::cpu_kernels::UnaryDerivative;
+use no_std_compat::f32::consts::PI;
```
coreylowman, line 2:
```suggestion
use std::f32::consts::PI;
```
since no_std_compat should overwrite the std namespace

coreylowman (commented):
Looks great! Small nits and then good to go. We can use this PR as a model for how to contribute new unary ops 🔥

src/tensor_ops/gelu/mod.rs
```diff
@@ -0,0 +1,67 @@
...
+        data.iter().map(|x| (x * m).round() / m).collect()
+    }
+
+    #[test]
+    fn test_gelu() {
+        let dev: TestDevice = Default::default();
+        let x = dev.tensor([-2.0, -1.0, 0.0, 1.0, 2.0]);
+        let r = x.trace().gelu();
+        assert_eq!(
+            simplify(&r.array()),
```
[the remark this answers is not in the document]
coreylowman: Ahh yeah looks like there's small failure in github actions, I think moving to assert_close should resolve that (probably don't need simplify)

Narsil:
The last failing test is the example I think, were I cannot use `assert_close` (afaik).

Should `assert_close` be opened ? Or maybe relax the assertion in a simple print statement ?

Narsil:
Why was the last test cancelled ??

coreylowman:
> Why was the last test cancelled ??

No idea, but the rest of them passed so looks good to me!

coreylowman:
> The last failing test is the example I think, were I cannot use `assert_close` (afaik).
> 
> Should `assert_close` be opened ? Or maybe relax the assertion in a simple print statement ?

Ah yeah, for doctests you don't need to call any assertions about data, they are more about just showing usage. You can just remove those lines

src/tensor_ops/gelu/mod.rs
```diff
@@ -0,0 +1,68 @@
...
+/// Examples:
+/// ```rust
+/// # use dfdx::prelude::*;
+/// # let dev: Cpu = Default::default();
+/// let t = dev.tensor([-1.0, 0.0, 1.0, 2.0]);
+/// let r = t.gelu();
+/// // assert_close(
+/// //     &r.array(),
+/// //     &[-0.04540229, -0.158808, 0.0, 0.841192, 1.9545977],
+/// // );
```
coreylowman, line 24:
```suggestion
```

coreylowman (approved):
Looks great from my end!

Narsil:
I'll let you merge in then ;) cheers !

coreylowman:
Nice contribution!

PR #398, opened 2023-01-24
Reorganize tensor_ops, and add cuda_utils.cuh

nkoppel:
I felt that tensor_ops was getting cluttered, so I categorized the ops based on how they're implemented and what they do. One edge case in this process was dropout, which I placed in `nn` with matmul, conv2d, and pool2d, because it is implemented differently than every other unary operation. This pr also adds `cuda_utils.cuh` to the tensor ops folder, makes the cuda kernels use it, and adds support for cuda header files to `build.rs`.


coreylowman (commented):
cuda header file & build.rs changes look awesome! left a separate comment about organization, let's discuss

nkoppel:
Yeah, I kind of agree that organizing the operations in that way would have been a detriment to new contributors and would make it more difficult to change implementations down the line. That organization was trying to make modifying the implementations of the binary and unary operators easier by grouping them together in one place. I'm now considering using preprocesser macros to further reduce the amount of boilerplate in the binary and unary cuda kernels. This should be much more effective at reducing the effort needed to make changes to these operations in the future, i.e. #363.

coreylowman (approved):
Hah wow, macros FTW! Love the number of deleted lines

coreylowman:
Ahh there's a conflict since #397 was just merged, hopefully easy fix! Then looks good to merge from me

nkoppel:
Ok, I've merged #397, and I intentionally left gelu.cu as is, because its complicated math doesn't work too well with my macro.

nkoppel:
I've now implemented a macro for binary operations.

coreylowman:
Awesome, nice work!

PR #405, opened 2023-01-26
Removes `ModuleBuilder`, Adds `BuildModule` & `BuildOnDevice`

### LINKED ISSUE #388: nn api hard to use with different devices
# Summary

Currently, a `Device` generic was added to all structs in `nn` that require storing tensors. E.g. `Linear<5, 10, Cpu>` or `Linear<5, 10, Cuda>`.

For an example of the current api for specifying a simple structure, you can specify:

```rust
type MLP = (Linear<5, 10>, ReLU, Linear<10, 1>);
```

However this currently defaults to the Cpu Device. In order to use this structure on a Cuda device, you'd have to go through and modify everything to set the new device generic:

```rust
type MLP<D: DeviceStorage> = (Linear<5, 10, D>, ReLU, Linear<10, 1, D>);
```

Additionally, when using with `dev.build_module()`, you have to specify the device twice here:

```rust
let dev: D = Default::default();
let m: MLP<D> = dev.build_module();
```

Even though the device should be identical to whatever device is building the module.

Further, if you happen to change the type of `dev`, you can get really hard to understand compile errors if you don't update the type you specified on the MLP:

```rust
let dev: D1 = Default::default();
// really weird errors because D1 can't build modules for device type D2
let m: MLP<D2> = dev.build_module();
```

# Ways forward

I see a few ways forward, each with their own advantages/disadvantages. There are probably more ways forward as well.

## Add a 2nd layer of nn configuration that would be used to build modules on devices

This would look something like this:

```rust
struct Linear<const I: usize, const O: usize>;

struct DeviceLinear<const I: usize, const O: usize, D: DeviceStorage> {
    weight: Tensor<Rank2<I, O>, f32, D>,
    bias: Tensor<Rank1<B>, f32, D>,
}

type MLP = (Linear<5, 10>, ReLU, Linear<10, 1>);

fn main() {
    let cpu: Cpu = Default::default();
    let cuda: Cuda = Default::default();

    let mlp_on_cpu: (DeviceLinear<5, 10, Cpu>, ReLU, DeviceLinear<5, 10, Cpu>) = dev.build_module<MLP>();
    let mlp_on_gpu: (DeviceLinear<5, 10, Cuda>, ReLU, DeviceLinear<5, 10, Cuda>) = dev.build_module<MLP>();
}
```
Specifically, the existing structures would be repurposes to concrete types for already build modules. The types `Linear<5, 10>` or `Conv2D<3, 2, 1>` would turn into almost configuration for device building.

Building device would tie these configuration structs together with the concrete build types:

```rust
trait BuildOn<D: DeviceStorage> {
    type Built;
}

impl BuildOn<D: DeviceStorage> for Linear<I, O> {
    type Built= DeviceLinear<I, O, D>;
}
```

### Pros

1. This makes specifying structure simple, actually it would be the exact same as you specify them now
2. You can re-use the same structure for different devices without changing the definition

### Cons

1. The constructed type is different from the type you use to specify the structure. So notably you have to specify the structure as a function generic instead of the output generic.
2. Having both DeviceLinear and Linear would be confusing, and we'd need documentation on what the difference is and when to use it
3. It is not obvious why this was done

## Add a `.to(device)` method of modules

This would mean you'd aways construct modules on Cpu and then send them to whatever device you wanted:

```rust
type MLP = (Linear<5, 10>, ReLU, Linear<10, 1>);

fn main() {
    let cpu: Cpu = Default::default();
    let cuda: Cuda = Default::default();

    let mlp_on_cpu: MLP = cpu.build_module();
    let mlp_on_gpu: (Linear<5, 10, Cuda>, ReLU, Linear<5, 10, Cuda>) = mlp_on_cpu.to(cuda);
}
```

### Pros

1. Existing structure would remain the same
2. `.to(device)` already exists in pytorch so people would be familiar with it
3. Very little documentation would need to be added, and this approach is probably easy to understand
4. Easy to implement with minimal changes

### Cons

1. This doesn't address the issues of having an additional device generic on the structures. For things like Conv2D or Transformer, you will have to specify all the parameters, even those with default values that you want to keep.
2. This still has the same issue of if you try to build with a different device than the type does.

coreylowman:
Resolves #388 

See discussions in issue and in #394.

This removes `ModuleBuilder`, notably you can't call `dev.build_module()` anymore. Instead users should use either `BuildModule` if they are using only the Cpu device, or if they want to specify device in the typing. Or the more general `BuildOnDevice`, which should be called like `Model::build_on_device(&dev)`.

A con of this is that this does complicate creating modules a bit since there are two distinct ways to create them. The main suggested pathway should be:

1. Specifying your model as a type alias (e.g. `type Model = ...`)
2. Using `BuildOnDevice` to build it (e.g. `let mut model = Model::build_on_device(&dev)`)

This makes it trivial to switch devices in the future.

# Error message for BuildOnDevice with the wrong device

```rust
error[E0308]: mismatched types
  --> examples/tmp.rs:6:43
   |
6  |     let m: Model = Model::build_on_device(&dev);
   |                    ---------------------- ^^^^ expected struct `dfdx::tensor::Cpu`, found struct `dfdx::tensor::Cuda`
   |                    |
   |                    arguments to this function are incorrect
   |
   = note: expected reference `&dfdx::tensor::Cpu`
              found reference `&dfdx::tensor::Cuda`
note: associated function defined here
  --> /home/ubuntu/dfdx/src/nn/module.rs:49:8
   |
49 |     fn build_on_device(device: &D) -> Self::Output {
   |        ^^^^^^^^^^^^^^^
```

# Error message for BuildModule with the wrong device

```rust
error[E0277]: the trait bound `dfdx::nn::Linear<5, 2>: dfdx::nn::BuildModule<dfdx::tensor::Cuda, f32>` is not satisfied
 --> examples/tmp.rs:5:39
  |
5 |     let m: Model = BuildModule::build(&dev);
  |                    ------------------ ^^^^ the trait `dfdx::nn::BuildModule<dfdx::tensor::Cuda, f32>` is not implemented for `dfdx::nn::Linear<5, 2>`
  |                    |
  |                    required by a bound introduced by this call
  |
  = help: the trait `dfdx::nn::BuildModule<D, f32>` is implemented for `dfdx::nn::Linear<I, O, D>`

For more information about this error, try `rustc --explain E0277`.
```

### DIFF
--- a/examples/06-mnist.rs
+++ b/examples/06-mnist.rs
@@ -99,7 +99,7 @@ fn main() {
 
     // initialize model and optimizer
     let mut model = Mlp::build_on_device(&dev);
-    let mut opt: Adam<Mlp> = Default::default();
+    let mut opt = Adam::default();
 
     // initialize dataset
     let dataset = MnistDataset::train(&mnist_path);
--- a/examples/06-mnist.rs
+++ b/examples/06-mnist.rs
@@ -74,10 +74,10 @@ impl MnistDataset {
 
 // our network structure
 type Mlp = (
-    (Linear<784, 512, Dev>, ReLU),
-    (Linear<512, 128, Dev>, ReLU),
-    (Linear<128, 32, Dev>, ReLU),
-    Linear<32, 10, Dev>,
+    (Linear<784, 512>, ReLU),
+    (Linear<512, 128>, ReLU),
+    (Linear<128, 32>, ReLU),
+    Linear<32, 10>,
 );
 
 // training batch size
@@ -98,7 +98,7 @@ fn main() {
     let mut rng = StdRng::seed_from_u64(0);
 
     // initialize model and optimizer
-    let mut model: Mlp = dev.build_module();
+    let mut model = Mlp::build_on_device(&dev);
     let mut opt: Adam<Mlp> = Default::default();
 
     // initialize dataset
--- a/src/nn/module.rs
+++ b/src/nn/module.rs
@@ -42,14 +42,20 @@ pub trait BuildModule<D: Device<E>, E: Dtype>: Sized {
 /// than it is on. Builds [BuildOnDevice::Built].
 ///
 /// Related to [BuildModule]
-pub trait BuildOnDevice<D: Device<E>, E: Dtype> {
-    type Built: BuildModule<D, E>;
-    fn build_on_device(device: &D) -> Self::Built {
+pub trait BuildOnDevice<D: Device<E>, E: Dtype>: ToDevice<D>
+where
+    Self::Output: BuildModule<D, E>,
+{
+    fn build_on_device(device: &D) -> Self::Output {
         BuildModule::build(device)
     }
-    fn try_build_on_device(device: &D) -> Result<Self::Built, D::Err> {
+    fn try_build_on_device(device: &D) -> Result<Self::Output, D::Err> {
         BuildModule::try_build(device)
     }
+}
+impl<T: ToDevice<D>, D: Device<E>, E: Dtype> BuildOnDevice<D, E> for T where
+    T::Output: BuildModule<D, E>
+{
 }
 
 /// Something that can reset it's parameters.
@@ -69,19 +75,26 @@ pub trait ResetParams<D: Device<E>, E: Dtype>: Sized {
 /// blanket impls for [ResetParams], [GradientUpdate], and [ModuleMut]
 pub trait ZeroSizedModule: Default {}
 
-impl<T: ZeroSizedModule + Clone, D: Device<E>, E: Dtype> ResetParams<D, E> for T {
+impl<T: ZeroSizedModule, D: Device<E>, E: Dtype> BuildModule<D, E> for T {
     fn try_build(_: &D) -> Result<Self, <D>::Err> {
         Ok(Default::default())
     }
 }
-impl<T: ZeroSizedModule, D: Device<E>, E: Dtype> BuildOnDevice<D, E> for T {
-    type Built = Self;
-}
+
 impl<T: ZeroSizedModule, D: Device<E>, E: Dtype> ResetParams<D, E> for T {
     fn try_reset_params(&mut self) -> Result<(), <D>::Err> {
         Ok(())
     }
 }
+
+impl<T: ZeroSizedModule + Clone, D> ToDevice<D> for T {
+    type Output = T;
+
+    fn to_device(&self, _device: &D) -> Self {
+        self.clone()
+    }
+}
+
 impl<T: ZeroSizedModule, D: Device<E>, E: Dtype> GradientUpdate<D, E> for T {
     fn update<U>(&mut self, _: &mut U, _: &mut crate::optim::UnusedTensors) -> Result<(), <D>::Err>
     where
@@ -102,13 +115,5 @@ where
     type Output = <Self as Module<T>>::Output;
     fn forward_mut(&mut self, input: T) -> Self::Output {
         self.forward(input)
-    }
-}
-
-impl<T: ZeroSizedModule + Clone, D> ToDevice<D> for T {
-    type Output = T;
-
-    fn to_device(&self, _device: &D) -> Self {
-        self.clone()
     }
 }
--- a/src/nn/module.rs
+++ b/src/nn/module.rs
@@ -24,15 +24,32 @@ pub trait ModuleMut<Input> {
     fn forward_mut(&mut self, input: Input) -> Self::Output;
 }
 
-/// Something that can reset it's parameters.
-pub trait ResetParams<D: Device<E>, E: Dtype>: Sized {
+/// Something that can be built. Related to [BuildOnDevice]
+pub trait BuildModule<D: Device<E>, E: Dtype>: Sized {
     /// Construct it on the device
     fn build(device: &D) -> Self {
         Self::try_build(device).unwrap()
     }
     /// Fallible version of [ResetParams::build]
     fn try_build(device: &D) -> Result<Self, D::Err>;
+}
 
+/// Something that can be built on a different device
+/// than it is on. Builds [BuildOnDevice::Built].
+///
+/// Related to [BuildModule]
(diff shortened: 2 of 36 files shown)


src/nn/module.rs
```diff
@@ -24,15 +24,32 @@ pub trait ModuleMut<Input> {
...
+/// Related to [BuildModule]
+pub trait BuildOnDevice<D: Device<E>, E: Dtype> {
+    type Built: BuildModule<D, E>;
+    fn build_on_device(device: &D) -> Self::Built {
+        BuildModule::build(device)
+    }
+    fn try_build_on_device(device: &D) -> Result<Self::Built, D::Err> {
+        BuildModule::try_build(device)
+    }
+}
```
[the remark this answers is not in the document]
coreylowman: Yep good point!

src/nn/module.rs
```diff
@@ -28,15 +28,38 @@ pub trait ModuleMut<Input> {
...
+        BuildModule::build(device)
+    }
+    fn try_build_on_device(device: &D) -> Result<Self::Output, D::Err> {
+        BuildModule::try_build(device)
+    }
+}
+impl<T: ToDevice<D>, D: Device<E>, E: Dtype> BuildOnDevice<D, E> for T where
+    T::Output: BuildModule<D, E>
+{
+}
```
coreylowman, line 59: This is currently causing super weird error messages.

Source:
```rust
use dfdx::prelude::*;
fn main() {
    type Model = Linear<5, 2, Cuda>;
    let dev: Cpu = Default::default();
    let m: Model = BuildModule::build(&dev);
}
```

error:
```rust
error[E0277]: the trait bound `dfdx::nn::Linear<5, 2, dfdx::tensor::Cuda>: ZeroSizedModule` is not satisfied
 --> examples/tmp.rs:8:39
  |
8 |     let m: Model = BuildModule::build(&dev);
  |                    ------------------ ^^^^ the trait `ZeroSizedModule` is not implemented for `dfdx::nn::Linear<5, 2, dfdx::tensor::Cuda>`
  |                    |
  |                    required by a bound introduced by this call
  |
  = help: the following other types implement trait `ZeroSizedModule`:
            Abs
            AvgPool2D<K, S, P>
            AvgPoolGlobal
            Cos
            Dropout
            DropoutOneIn<N>
            Exp
            Flatten2D
          and 13 others
  = note: required for `dfdx::nn::Linear<5, 2, dfdx::tensor::Cuda>` to implement `dfdx::nn::BuildModule<dfdx::tensor::Cpu, f32>`
```

src/nn/module.rs
```diff
@@ -28,15 +28,38 @@ pub trait ModuleMut<Input> {
     fn forward_mut(&mut self, input: Input) -> Self::Output;
 }
 
-/// Something that can reset it's parameters.
-pub trait ResetParams<D: Device<E>, E: Dtype>: Sized {
+/// Something that can be built. Related to [BuildOnDevice]
+pub trait BuildModule<D: Device<E>, E: Dtype>: Sized {
```
coreylowman, line 32: Something is causing really weird errors with this:

Source:
```rust
use dfdx::prelude::*;
fn main() {
    type Model = Linear<5, 2, Cuda>;
    let dev: Cpu = Default::default();
    let m: Model = BuildModule::build(&dev);
}

```

error:
```rust
error[E0277]: the trait bound `dfdx::nn::Linear<5, 2, dfdx::tensor::Cuda>: ZeroSizedModule` is not satisfied
 --> examples/tmp.rs:8:39
  |
8 |     let m: Model = BuildModule::build(&dev);
  |                    ------------------ ^^^^ the trait `ZeroSizedModule` is not implemented for `dfdx::nn::Linear<5, 2, dfdx::tensor::Cuda>`
  |                    |
  |                    required by a bound introduced by this call
  |
  = help: the following other types implement trait `ZeroSizedModule`:
            Abs
            AvgPool2D<K, S, P>
            AvgPoolGlobal
            Cos
            Dropout
            DropoutOneIn<N>
            Exp
            Flatten2D
          and 13 others
  = note: required for `dfdx::nn::Linear<5, 2, dfdx::tensor::Cuda>` to implement `dfdx::nn::BuildModule<dfdx::tensor::Cpu, f32>`
```

coreylowman: Removing the blanket impl of ZeroSizedModule results in the following error:

```rust
error[E0277]: the trait bound `dfdx::nn::Linear<5, 2>: dfdx::nn::BuildModule<dfdx::tensor::Cuda, f32>` is not satisfied
 --> examples/tmp.rs:5:39
  |
5 |     let m: Model = BuildModule::build(&dev);
  |                    ------------------ ^^^^ the trait `dfdx::nn::BuildModule<dfdx::tensor::Cuda, f32>` is not implemented for `dfdx::nn::Linear<5, 2>`
  |                    |
  |                    required by a bound introduced by this call
  |
  = help: the trait `dfdx::nn::BuildModule<D, f32>` is implemented for `dfdx::nn::Linear<I, O, D>`
```

src/nn/module.rs
```diff
@@ -28,15 +28,38 @@ pub trait ModuleMut<Input> {
...
     /// Construct it on the device
     fn build(device: &D) -> Self {
         Self::try_build(device).unwrap()
     }
     /// Fallible version of [ResetParams::build]
```
coreylowman, line 37:
```suggestion
    /// Fallible version of [BuildModule::build]
```

examples/06-mnist.rs
```diff
@@ -98,8 +98,8 @@ fn main() {
     let mut rng = StdRng::seed_from_u64(0);
 
     // initialize model and optimizer
-    let mut model: Mlp = dev.build_module();
-    let mut opt: Adam<Mlp> = Default::default();
```
coreylowman, line 102: Users probably shouldn't specify `Adam<Mlp>` anymore, since that will change based on device. We'll probably need to not allow Default anymore and pass in what you want to optimize to the optimizer constructor. This should throw a helpful error if you pass in the wrong thing.

PR #406, opened 2023-01-26
Adding nn.Embedding layer.

Narsil:
Attempt to create `nn.Embedding` layer.

However I am not able to finish the gradients part.


src/nn/embedding.rs
```diff
@@ -0,0 +1,206 @@
...
+    /// Transposed weight matrix, shape (I, O)
+    pub weight: Tensor<Rank2<VOCAB, DIM>, f32, D>,
+}
+
+impl<const VOCAB: usize, const DIM: usize, const SEQ: usize, D: Device<f32>>
+    Module<Tensor<Rank1<SEQ>, usize, D>> for Embedding<VOCAB, DIM, D>
+{
+    type Output = Tensor<Rank2<SEQ, DIM>, f32, D>;
+    fn forward(&self, input: Tensor<Rank1<SEQ>, usize, D>) -> Self::Output {
+        self.weight.retaped().gather(input)
```
Narsil, line 40: This is the particular line that seems to be not working.

In other layer I found

```rust
self.weight.retaped::<T>()  // T is the tape of the input 
```
Which seems to be the trick. However, it seems to me that the weight could contain the tape, and the input cannot ( since it's only indexing within the weight tensor).

I'm out of ideas to make this work. Could you provide any help @coreylowman  ?

Narsil: I figure out a way by making `GatherTo` generic on the Tape too. I'm not to sure about the modifications though.

src/tensor_ops/select_and_gather/mod.rs
```diff
@@ -157,9 +160,9 @@ pub trait GatherTo<D: DeviceStorage>: HasErr + HasShape {
 impl<Src: Shape, E: Dtype, D: ReplaceDimKernel<E>, T: Tape<D>> GatherTo<D>
     for Tensor<Src, E, D, T>
 {
-    fn try_gather<Dst: Shape, Idx: Shape>(
+    fn try_gather<Dst: Shape, Idx: Shape, TT: Tape<D>>(
```
coreylowman, line 163: typo? 😀

src/nn/embedding.rs
```diff
@@ -0,0 +1,230 @@
...
+        const DIM: usize,
+        const SEQ: usize,
+        const BATCH: usize,
+        D: Device<f32>,
+        T: Tape<D>,
+    > Module<Tensor<Rank2<BATCH, SEQ>, usize, D, T>> for Embedding<VOCAB, DIM, D>
+{
+    type Output = Tensor<Rank3<BATCH, SEQ, DIM>, f32, D, T>;
+    fn forward(&self, input: Tensor<Rank2<BATCH, SEQ>, usize, D, T>) -> Self::Output {
+        self.weight.retaped::<T>().gather(input)
```
coreylowman, line 55: One trick we could do here is with SplitTape and PutTape:

```suggestion
    fn forward(&self, input: Tensor<Rank2<BATCH, SEQ>, usize, D, T>) -> Self::Output {
        let (input, tape) = input.split_tape();
        self.weight.clone().put_tape(tape).gather(input)
```

I think this should avoid the need to change select/gather?

coreylowman, line 55: Also, do you still need the retaped even with the modifications to gather?

src/tensor_ops/select_and_gather/mod.rs
Narsil: No theres another T being used by `impl`.

src/nn/embedding.rs
Narsil: It works ! 
Still somehow mystical how the tape thing works :)

Narsil: Not with the manual split_tape.

coreylowman: Nice! Hmm I wonder what we could to do make it easier to understand intuitively... That is a big difference with pytorch, and while most use cases shouldn't need to do stuff with tapes, for understanding internals it would be helpful.

src/tensor_ops/select_and_gather/mod.rs
coreylowman: Oooooh good call, missed that

coreylowman (approved):
Looks great, thanks for the contribution!

PR #409, opened 2023-01-26
Fix GELU CUDA kernel compilation

### LINKED ISSUE #407: GELU cuda kernel does not compile, missing header
## Problem

When trying to compile I get this `nvcc` error:

```
nvcc error while compiling "src\tensor_ops\gelu\gelu.cu": 
Output {
    status: ExitStatus(ExitStatus(1)),
    stdout: "c:\Users\vilia\Code\dfdx\src\tensor_ops\gelu\gelu.cu(5): error: identifier "M_2_SQRTPI" is undefined
        c:\Users\vilia\Code\dfdx\src\tensor_ops\gelu\gelu.cu(5): error: identifier "M_SQRT1_2" is undefined
        c:\Users\vilia\Code\dfdx\src\tensor_ops\gelu\gelu.cu(5): error: identifier "M_2_SQRTPI" is undefined
        c:\Users\vilia\Code\dfdx\src\tensor_ops\gelu\gelu.cu(5): error: identifier "M_SQRT2" is undefined
        4 errors detected in the compilation of "src/tensor_ops/gelu/gelu.cu". gelu.cu",
    stderr: ""
}
```
(formatting mine)

## Info

`rustc --version --verbose`
```
rustc 1.66.0 (69f9c33d7 2022-12-12)
binary: rustc
commit-hash: 69f9c33d71c871fc16ac445211281c6e7a340943
commit-date: 2022-12-12
host: x86_64-pc-windows-msvc
release: 1.66.0
LLVM version: 15.0.2
```

`nvcc --version`
```
nvcc: NVIDIA (R) Cuda compiler driver
Copyright (c) 2005-2022 NVIDIA Corporation
Built on Mon_Oct_24_19:40:05_Pacific_Daylight_Time_2022
Cuda compilation tools, release 12.0, V12.0.76
Build cuda_12.0.r12.0/compiler.31968024_0
```

## Cause?

These constants are defined in `math.h` or `cmath`. I believe they are not linked by default, at least in [MSVC](https://learn.microsoft.com/en-us/cpp/c-runtime-library/math-constants?view=msvc-170) and [GCC](https://learn.microsoft.com/en-us/cpp/c-runtime-library/math-constants?view=msvc-170). For [GNU](https://www.gnu.org/software/libc/manual/html_node/Mathematical-Constants.html) they should be (I think?), but when I switched to the `stable-x86_64-pc-windows-gnu` toolchain I was still unable to compile (same error).

ViliamVadocz:
Closes #407 

Math Constant macros were not included which caused issues when trying to compile.
While at it, I also changed `tanh` to `tanhf` and added a `constexpr`  where appropriate.

### DIFF
--- a/src/tensor_ops/gelu/gelu.cu
+++ b/src/tensor_ops/gelu/gelu.cu
@@ -1,4 +1,6 @@
 #include "unary_op_macros.cuh"
+#define _USE_MATH_DEFINES
+#include <math.h>
 
 struct GeLUKernelOp {};
 
@@ -10,16 +12,16 @@ LONG_UNARY_OP(gelu_forward, gelu_backward, GeLUKernelOp,
 
         float alpha = x + fastCoeff * x_cube;
 
-        float y = 0.5 * x * (1.0 + tanh(M_2_SQRTPI * M_SQRT1_2 * alpha));
+        float y = 0.5 * x * (1.0 + tanhf(M_2_SQRTPI * M_SQRT1_2 * alpha));
         out[i] = y;
     },
     {
-        float kBeta = M_2_SQRTPI * M_SQRT2 * 0.5;                       
+        constexpr float kBeta = M_2_SQRTPI * M_SQRT2 * 0.5;                       
         constexpr float fastCoeff = 0.044715;
         float x_sq = x * x;
         float x_cube = x_sq * x;
         float inner = kBeta * (x + fastCoeff * x_cube);
-        float tanh_inner = tanh(inner);
+        float tanh_inner = tanhf(inner);
 
         float left = 0.5 * x;
         float right = 1.0 + tanh_inner;


src/tensor_ops/gelu/gelu.cu
```diff
@@ -1,4 +1,6 @@
 #include "unary_op_macros.cuh"
+#define _USE_MATH_DEFINES
```
coreylowman, line 2: What does this do exactly? Also have you tried compiling all this on a non GNU and MSVC device?

ViliamVadocz: From the [MSVC documentation](https://learn.microsoft.com/en-us/cpp/c-runtime-library/math-constants?view=msvc-170)
> The math constants aren't defined in Standard C/C++. To use them, you must first define `_USE_MATH_DEFINES`, and then include `<cmath>` or `<math.h>`.

The `#define` is used here to essentially set a flag so that the math constants are defined in the `math.h` header file. 
If we take a look at the ([relevant documentation](https://learn.microsoft.com/en-us/cpp/preprocessor/hash-define-directive-c-cpp?source=recommendations&view=msvc-170)) we can see how it works:

> A `#define` without a token-string removes occurrences of identifier from the source file. The identifier remains defined and **can be tested** by using the `#if defined` and `#ifdef` directives.

(Emphasis mine)

---

I have not tested with other toolchains than `stable-x86_64-pc-windows-gnu` and `stable-x86_64-pc-windows-msvc`.

ViliamVadocz:
Apparently there exist also floating point versions of these constants which are available if `_GNU_SOURCE` is also defined. The constants would be appended with a lowercase `f` to get the floating point version. Info: https://www.gnu.org/software/libc/manual/html_node/Mathematical-Constants.html

PR #412, opened 2023-01-26
[Proposal] Adds AutoDevice

### LINKED ISSUE #410: [Ideation - Don't work on] Add DefaultDevice depending on feature flags - use in Tensor & all nn modules
This has the potential to address #408 #388.

Potential to revert both #394, #405.

In many applications I'd expect we'd see something like this (and you can already see this in examples/06-mnist.rs:
```rust
#[cfg(not(feature = "cuda"))]
type Dev = Cpu;

#[cfg(feature = "cuda")]
type Dev = Cuda;
```

The idea of this issue is to add this to dfdx and then change both tensor & nn structs to use this:

```rust
struct Tensor<..., D: DeviceStorage = DefaultDevice> { ... }

struct Linear<..., D: DeviceStorage = DefaultDevice> { ... }
// etc
```

We can even go one step further and add another feature flag to override the default device, to make it settable by user.

# What this solves

- This makes it so you don't have to change any code to start using cuda, only turn on the feature flag
- This makes it so you shouldn't ever have to specify the device generic.
- This could make BuildOnDevice & ToDevice irrelevant (probably useful in some situations, but not worth keeping for now)

coreylowman:
Resolves #410

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,8 +66,7 @@ type Mlp = (
 
 fn main() {
     let dev: Cpu = Default::default();
-    // OR `let dev: Cuda = Default::default();`
-    let mlp = Mlp::build_on_device(&dev);
+    let mlp: Mlp = dev.build_module();
     let x: Tensor<Rank1<10>> = dev.zeros();
     let y /*: Tensor<Rank1<2>>*/ = mlp.forward(x);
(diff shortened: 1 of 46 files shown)


coreylowman:
Closing for more thinking

Issue #399, opened 2023-01-25
Create tensors directly from vec's easily


coreylowman:
Right now you have to create the tensor first, then copy into the tensor from a vec. It'd be nice to just create a tensor directly from a vec. The underlying devices make this easy with `take_async` for cuda and you can already create a StridedArray with a vec with Cpu.

The main challenge will be figuring out a nice interface for both const/dyn shapes. I.e. for dyn shapes, the user needs to be able to specify the shape of the tensor, but that is not needed for const.

Also, this function/method should return a `Result<Tensor<...>, Vec<E>>`, in the case that the number of elements in the vec were not enough.

nkoppel:
Here's what I've been drafting for this:
```rust
pub trait TensorFrom<S: Shape, Src, E: Unit>: DeviceStorage {
    fn tensor_with_shape(&self, src: Src, shape: &S) -> Tensor<S, E, Self> {
        self.try_tensor_with_shape(src, shape).unwrap()
    }

    fn try_tensor_with_shape(&self, src: Src, shape: &S) -> Result<Tensor<S, E, Self>, Self::Err>;
}

pub trait ConstShapeTensorFrom<S: ConstShape, Src, E: Unit>: DeviceStorage {
    fn tensor(&self, src: Src) -> Tensor<S, E, Self> {
        self.try_tensor(src).unwrap()
    }

    fn try_tensor(&self, src: Src) -> Result<Tensor<S, E, Self>, Self::Err>;
}

impl<S: ConstShape, Src, E: Unit, T: TensorFrom<S, Src, E>> ConstShapeTensorFrom<S, Src, E> for T {
    fn try_tensor(&self, src: Src) -> Result<Tensor<S, E, Self>, Self::Err> {
        self.try_tensor_with_shape(src, &S::default())
    }
}
```
This generalizes TensorFromArray, allowing `device.tensor()` to be polymorphic over arrays, vecs, iterators, etc.

coreylowman:
Closed from PR above

PR #414, opened 2023-01-28
Add TensorFrom trait to create tensors from both vectors and arrays.

nkoppel:
Draft of a solution for #399, which splits TensorFromArray into ConstShapeTensorFrom, DynamicTensorFrom, and the public-facing TensorFrom. Currently, this doesn't work how I'd like it to, as this cannot be used when the device is a generic`Device` because the traits need to be parameterized by a Shape, which means that it can't be added to the definition of `Device`.


ViliamVadocz:
Would it be possible to have stack-allocated tensors if they're small enough?

nkoppel:
Currently, no, because StridedArrays store tensors as an `Arc<Vec<T>>`, which is always stored on the heap, so using something like [smallvec](https://docs.rs/smallvec/latest/smallvec/) wouldn't have any performance benefit.

src/tensor/cpu/allocate.rs
```diff
@@ -197,6 +197,29 @@ impl<E: Unit, const M: usize, const N: usize, const O: usize, const P: usize>
...
 
+impl<E: Unit> TensorFromVec<E> for Cpu {
+    fn try_tensor_from_vec<S: Shape>(
+        &self,
+        mut src: Vec<E>,
+        shape: S,
+    ) -> Result<Tensor<S, E, Self>, Self::Err> {
+        let num_elements = shape.num_elements();
+
+        if src.len() < num_elements {
```
coreylowman, line 208: We should do this if there are too many elements as well - the truncation is a too silent in my opinion.

src/tensor/cuda/allocate.rs
```diff
@@ -138,6 +138,16 @@ where
     }
 }
 
+impl<E: Unit> TensorFromVec<E> for Cuda {
+    fn try_tensor_with_shape<S: Shape>(
+        &self,
+        src: Vec<E>,
+        shape: S,
+    ) -> Result<Tensor<S, E, Self>, Self::Err> {
+        self.take_cpu_tensor(self.cpu.try_tensor_with_shape(src, shape)?)
```
coreylowman, line 147: Let's fill in this method directly instead of going through cpu here. Mainly because we can use self.dev.take_async(src) directly since we have the vec, which will be much faster than going through the cpu!

It's fine to duplicate error handling here IMO.

coreylowman (changes requested):
Looks great! Want to also update the examples/06-mnist to use this new trait?

examples/06-mnist.rs
```diff
@@ -64,11 +64,7 @@ impl MnistDataset {
             choices[self.lbl[img_idx]] = 1.0;
             lbl_data.extend(choices);
         }
-        let mut img = dev.zeros();
-        img.copy_from(&img_data);
-        let mut lbl = dev.zeros();
-        lbl.copy_from(&lbl_data);
-        (img, lbl)
+        (dev.tensor_from_vec(img_data), dev.tensor_from_vec(lbl_data))
```
coreylowman, line 67: 🎉

src/tensor/storage_traits.rs
```diff
@@ -249,6 +250,44 @@ pub trait TensorFromArray<Src, S: Shape, E: Unit>: DeviceStorage {
...
 
+/// Construct tensors from rust vectors
+pub trait TensorFromVec<E: Unit>: DeviceStorage {
+    /// Create a tensor with a dynamic shape from a rust vector
+    /// ```rust
+    /// # use dfdx::prelude::*;
+    /// # let dev: Cpu = Default::default();
+    /// let _ = dev.dynamic_tensor_from_vec(vec![1.0, 2.0, 3.0, 4.0, 5.0, 6.0], [2, 3]);
+    /// ```
+    fn dynamic_tensor_from_vec<S: Shape>(&self, src: Vec<E>, shape: S) -> Tensor<S, E, Self> {
```
coreylowman, line 261: Thoughts on these names?
- `tensor_from_vec_with_shape`
- `tensor_from_shaped_vec`
- `tensor_from_vec_like` (similar to `zeros_like`/`ones_like`/etc, however I do think that is less clear in this case)

I think the `with_shape` one is the most clear about how its different from the other calls of this trait, and how you call it. Usage of `dynamic` doesn't necessarily imply you have to give a shape, but I do like that it implies it should be used with runtime shapes.

coreylowman (commented):
Looking great! Just need to settle on a good name for the non-const shape method

src/tensor/storage_traits.rs
nkoppel: Yeah, I've gone with tensor_from_vec_with_shape, but I'm not super happy that the names are as long as they are. It's kind of a shame that rust doesn't seem to have the features necessary to have the features necessary for us to merge the TensorFromVec and TensorFromArray traits while still allowing users to use these traits with a generic `Device<E>`.

nkoppel: Yes, we could do this, but doing so would make it impossible to use dev.tensor when using a generic Device, because the extra shape parameter makes it impossible to add as a subtrait of Device. Also, that code snippet would have to look something like:
```rust
let a: Tensor<Rank1<3>, ...> = dev.tensor([1.0, 2.0, 3.0]);
let a: Tensor<Rank1<3>, ...> = dev.tensor(vec![1.0, 2, 3]);
let a: Tensor<(usize, ), ...> = dev.tensor_with_shape((vec![1.0, 2.0, 3.0], (3, ));
```
to allow optionally passing in a shape.

I may be missing something, but I was unable to merge TensorFromArray and TensorFromVec while still having the resulting trait be a subtrait of Device.

nkoppel:
Per your previous comment, I've greatly simplified the api, and it is now possible to create tensors from arrays and vectors, even with a generic Device, with only the `tensor` method on a device. To do this, I've kept the TensorFromVec trait as an internal subtrait of TensorFrom to allow creating tensors on anything that implements DeviceStorage and TensorFromVec.

coreylowman (approved):
Nice this looks great! Doc example is very clear 🚀

Issue #305, opened 2022-12-11
Add tensor_ops::where to replace value_mask


coreylowman:
Should roughly look like:

```rust
fn where<S: Shape, E: Dtype, D: Device<E>>(x: Tensor<S, E, D>, condition: Tensor<S, bool, D> , y: Tensor<S, E, D>) -> Tensor<S, E, D>;
```

Whichever is the self tensor for method version should be the one with the tape of the output.

ux:
```rust
let gt0 = x.clone().gt(0);
x.where(gt0, y);
```

nkoppel:
Since 'where' is a keyword, I think I'm going to rename this operation to choose.

PR #415, opened 2023-01-29
Add boolean operations and choose.

nkoppel:
Implements `and`, `or`, `xor`, and `not` for Tensors. In the process, however, I expanded the definition of OnesTensor and OneFillTensor to all implementors of Unit, which caused Cpu and Cuda to implement Device for all implementors of Unit. Because of this, we now have to specify tensors' dtypes more often.

I also had to use take_async instead of alloc_zeros_async to create the result tensors because cudarc doesn't currently implement ValidAsZeroBits for booleans. This is already fixed in cudarc's latest main, so this can be changed the next time cudarc releases. I figured that take_async was preferred to transmuting a CudaSlice\<u8\> to a CudaSlice\<bool\>.

Now also implements `choose`, resolving #305.


coreylowman:
Looking great on first read through - probably going to need a couple more days with this one 😀

src/tensor_ops/boolean/mod.rs
```diff
@@ -0,0 +1,320 @@
...
+}
+
+macro_rules! boolean_op_impl {
+    ($op:ident, $method1:ident, $method2:ident, $method3:ident) => {
+        impl<S: Shape, D: BooleanKernel> $op for Tensor<S, bool, D> {
+            type Output = Self;
+
+            fn $method1(self, rhs: Self) -> Self {
+                self.device
+                    .upgrade(self.device.$method2(&self.storage, &rhs.storage).unwrap())
```
coreylowman, line 100: Related to #420, but all of these need to assert that the shapes are equal for runtime shapes. `assert_eq!(self.shape(), rhs.shape()`

src/tensor_ops/choose/mod.rs
```diff
@@ -0,0 +1,147 @@
...
+    > ChooseFrom<Tensor<S, E, D, LhsTape>, Tensor<S, E, D, RhsTape>> for Tensor<S, bool, D>
+{
+    type Output = Tensor<S, E, D, LhsTape>;
+
+    fn try_choose(
+        self,
+        lhs: Tensor<S, E, D, LhsTape>,
+        rhs: Tensor<S, E, D, RhsTape>,
+    ) -> Result<Self::Output, Self::Err> {
+        let (lhs, tape) = lhs.split_tape();
```
coreylowman, line 58: Similar to other comment, but need to add an assert for shape equality for both pairs:
```rust
assert_eq!(self.shape(), lhs.shape());
assert_eq!(self.shape(), rhs.shape());
```

src/tensor_ops/choose/mod.rs
```diff
@@ -0,0 +1,147 @@
...
+        &self,
+        cond: &Self::Storage<S, bool>,
+        grad_lhs: &mut Self::Storage<S, E>,
+        grad_rhs: &mut Self::Storage<S, E>,
+        grad_out: &Self::Storage<S, E>,
+    ) -> Result<(), Self::Err>;
+}
+
+/// Choose values from two tensors using a boolean mask. Equivalent to `torch.where` from pytorch.
+pub trait ChooseFrom<Lhs, Rhs>: HasErr {
```
coreylowman, line 30: Our first trinary operation! 😀

src/tensor_ops/boolean/cuda_kernels.rs
```diff
@@ -0,0 +1,106 @@
...
+        if !self.dev.has_func(MODULE_NAME, fn_name) {
+            self.dev
+                .load_ptx(PTX_SRC.into(), MODULE_NAME, &ALL_FN_NAMES)?;
+        }
+
+        let shape = lhs.shape;
+        let strides = lhs.shape.strides();
+        let numel = shape.num_elements();
+
+        let mut storage = self.dev.take_async(std::vec![false; numel])?;
```
coreylowman, line 27: Ah so this is where we need the ValidAsZeroBits? Can you add a todo comment in here, and then after we merge this we can open an issue to track it

src/tensor_ops/boolean/cuda_kernels.rs
```diff
@@ -0,0 +1,106 @@
+use super::BooleanKernel;
+use crate::prelude::{cuda::CudaArray, *};
+use cudarc::prelude::*;
+
+use std::sync::Arc;
+
+const MODULE_NAME: &str = "/boolean.ptx";
```
coreylowman, line 7: nitpick for consistency
```suggestion
const MODULE_NAME: &str = "boolean";
```

src/tensor_ops/boolean/cuda_kernels.rs
```diff
@@ -0,0 +1,106 @@
...
+        &self,
+        inp: &Self::Storage<S, bool>,
+    ) -> Result<Self::Storage<S, bool>, Self::Err> {
+        if !self.dev.has_func(MODULE_NAME, "boolean_not") {
+            self.dev
+                .load_ptx(PTX_SRC.into(), MODULE_NAME, &ALL_FN_NAMES)?;
+        }
+
+        let numel = inp.data.len();
+        let mut storage = self.dev.take_async(std::vec![false; numel])?;
```
coreylowman, line 65: Same thing with a TODO comment here

src/tensor_ops/boolean/mod.rs
```diff
@@ -0,0 +1,320 @@
...
+    fn scalar_and<S: Shape>(
+        &self,
+        lhs: &Self::Storage<S, bool>,
+        rhs: bool,
+    ) -> Result<Self::Storage<S, bool>, Self::Err> {
+        if rhs {
+            Ok(lhs.clone())
+        } else {
+            Ok(self.try_zeros_like(lhs)?.storage)
+        }
```
coreylowman, line 49: I think all three of these scalar methods can be removed from this trait and implemented directly in the macro below (or separately). There's weirdness in using tensor methods, unwrapping them for the storage, and then re-wrapping them in the traits down below.

coreylowman (changes requested):
this is looking great! I really like the choose method, great idea! Added a few structural comments, but definitely getting close on this one, thanks for the work!

src/tensor_ops/boolean/mod.rs
```diff
@@ -0,0 +1,320 @@
...
+impl<S: Shape, D: BooleanKernel> Not for &Tensor<S, bool, D> {
+    type Output = Tensor<S, bool, D>;
+
+    fn not(self) -> Self::Output {
+        self.device.upgrade(self.device.not(&self.storage).unwrap())
+    }
+}
+
+macro_rules! boolean_op_impl {
+    ($op:ident, $method1:ident, $method2:ident, $method3:ident) => {
```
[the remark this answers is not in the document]
nkoppel: I've tweaked your suggestions here a little, but I've made these macro parameters have more distinct names.

coreylowman:
Thanks for the contribution! 🚀

PR #416, opened 2023-01-29
[WIP/RFC] Named dynamic dims: `Dyn<'B'>`.

Narsil:
This PR tries to get named dynamic dimensions into the library.

The rationale is that some dynamic code can get messy really quick with
`usize` when there are several dimensions which are dynamic.

Let's take a realistic example: an inference webserver doing dynamic
batching on a transformer model.

```rust
const DIM: usize = 768;
const H: usize = 12;
let batch : Tensor<(usize, usize, Const<DIM>)> = dev.zeros_like(&(batch,
sequence_length, Const));

/// Within the transformer's attention.
/// Equivalent of https://github.com/coreylowman/dfdx/blob/main/src/nn/transformer/mha.rs
/// with dynamic sizes

let v: Tensor<(usize, usize, DIM)>, _, _, _> = self.w_v.forward(v.retaped::<T>());
let v = v.reshape::<(usize, usize, Const<H>, Const<{ DIM / H }>)>();
let v = v.permute::<(usize, Const<H>, usize, Const<{ DIM / H }>)>();
```

And now this is where it hurts, it starts to be non obvious intent.
Did we mean to swap out axis0/axis2 ? or not ?

And this gets exacerbated when trying to work with Encoder/Decoder (two
sequence lengths) and with past_key_values (another sequence_length
within the decoder).

The goal is to replace (or simply add it) this with:

```rust
const DIM: usize = 768;
let batch : Tensor<(Dyn<'B'>, Dyn<'S'>, Const<DIM>)> = dev.zeros_like(&(batch,
sequence_length, Const));

/// Within the transformer's attention.
/// Equivalent of https://github.com/coreylowman/dfdx/blob/main/src/nn/transformer/mha.rs
/// with dynamic sizes

let v: Tensor<(Dyn<'B'>, Dyn<'S'>, Const<DIM>)>, _, _, _> = self.w_v.forward(v.retaped::<T>());
let v = v.reshape::<(Dyn<'B'>, Dyn<'S'>, Const<H>, Const<{ V / H }>)>();
let v = v.permute::<(Dyn<'B'>, Const<H>, Dyn<'S'>, Const<{ V / H }>)>();
```

### Pros

- More readable with multiple dynamic axes
- Less readable in simple cases
- Actual type safety:

```rust
let a : Tensor<Dyn<'B'>> = dev.zeros_like(&(Dyn::<'B'>(5),));
// Other part of the code
let b : Tensor<Dyn<'C'>> = dev.zeros_like(&(Dyn::<'C'>(5),));

let c = a + b; // Invalid
```

### Cons

- Might get tricky to represent to represent sums like Dyn<'B' + 1 > /
  Dyn<'B' / 2> (Is there a real need though ? ror transformers with past there is indeed a sequence addition (past + current) but if everything is dynamic it should be relatively doable)
- Might be confusing that's this size is **not** enforced at compile
  time (since it's dynamic). Only the name is enforced.
  So:

```rust
let dynamic: Tensor<(Dyn<'B'>, Const<3>, Const<4>)> =
    dev.ones_like(&(Dyn::<'B'>(5), Const, Const));
let dynamic2: Tensor<(Dyn<'B'>, Const<3>, Const<4>)> =
    dev.ones_like(&(Dyn::<'B'>(2), Const, Const));
let a = dynamic + dynamic2;
```

This is a runtime error (currently silently failing to do the right
computation). This was already the case, but this is even less obvious.

This is all I can come up with.
What do you think.

### DIFF
--- a/examples/01-tensor.rs
+++ b/examples/01-tensor.rs
@@ -1,7 +1,7 @@
 //! Intro to dfdx::arrays and dfdx::tensor
 
 use dfdx::{
-    shapes::{Const, HasShape, Rank1, Rank2, Rank3},
+    shapes::{Const, Dyn, HasShape, Rank1, Rank2, Rank3},
     tensor::{AsArray, Cpu, OnesTensor, SampleTensor, Tensor, TensorFromArray, ZerosTensor},
 };
 
@@ -24,7 +24,8 @@ fn main() {
     let _: Tensor<Rank3<1, 2, 3>, f32, Cpu> = dev.ones();
 
     // Dynamic size
-    let dynamic: Tensor<(usize, Const<3>, Const<4>), f32, Cpu> = dev.zeros_like(&(5, Const, Const));
+    let dynamic: Tensor<(Dyn<'B'>, Const<3>, Const<4>), f32, Cpu> =
+        dev.zeros_like(&(Dyn::<'B'>(5), Const, Const));
     println!("Dynamic shape={:?}", dynamic.shape());
 
     // each of the creation methods also supports specifying the shape on the function
--- a/examples/03-nn.rs
+++ b/examples/03-nn.rs
@@ -2,7 +2,7 @@
 
 use dfdx::{
     nn::{BuildOnDevice, Linear, Module, ModuleMut, ReLU, ResetParams},
-    shapes::{Const, Rank1, Rank2},
+    shapes::{Const, Dyn, Rank1, Rank2},
     tensor::{AsArray, Cpu, SampleTensor, Tensor, ZerosTensor},
 };
 
@@ -31,7 +31,8 @@ fn main() {
 
     // Even dynamic size is supported;
     let batch_size = 3;
-    let _: Tensor<(usize, Const<2>), f32, _> = m.forward(dev.zeros_like(&(batch_size, Const)));
+    let _: Tensor<(Dyn<'B'>, Const<2>), f32, _> =
+        m.forward(dev.zeros_like(&(Dyn::<'B'>(3), Const)));
 
     // you can also combine multiple modules with tuples
     type Mlp = (Linear<4, 2>, ReLU, Linear<2, 1>);
--- a/src/data.rs
+++ b/src/data.rs
@@ -4,7 +4,7 @@ use rand::prelude::SliceRandom;
 use std::vec::Vec;
 
 use crate::{
-    shapes::{Const, Rank1},
+    shapes::{Const, Dyn, Rank1},
     tensor::{CopySlice, DeviceStorage, Tensor, ZerosTensor},
 };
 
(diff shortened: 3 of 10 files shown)


Narsil:
@coreylowman

Narsil:
> I think this is the major concern for me. Very hesitant about something that cool fool people into thinking things should compile but won't actually.

I 100% agree, but dynamic shapes are what makes PyTorch, really practical in situations where you want something dynamic (like I have shown, mostly batch sizes and sequence lengths in NLP. But could be resolution in Stable Diffusion, and you don't want to recompile everything all the time (this is what makes XLA, so Tensorflow and Jax very impractical at inference time, at training time, usually it's ok and doable to be fully static).

What I think could work in the Rust world though, would be to make every op return result type  for things that are dynamic, and only enable the non result methods that are sure to never fail.
This does remove the "it compiles so it works" property (even though I think the potential for error is much much less than in regular pytorch, since all tensors shapes are still known in advance. So you can't introduce new ops that screw your dimensions internally.

Narsil:
> Just to add to the conversation, at least with reshape, you can use the 2nd generic arg instead of first to specify axes:

Yes, I don't necessarily find that easier to read. It is more explicit about what's being doing, but does put strain on the reader to actually figure out who's who. In the other case, you can simply read the new shape, and that's the new shape.

coreylowman:
> What I think could work in the Rust world though, would be to make every op return result type for things that are dynamic, and only enable the non result methods that are sure to never fail.

Totally agree, we can mirror the `checked_*` (like checked_add) from the std library and return an option. I believe someone floated this idea for adding checked_reshape so we can take it off nightly too.

coreylowman:
I had another idea to use lifetimes to generate unique types (that can still be cloned). See:
- https://docs.rs/unique-type/latest/unique_type/
- https://www.reddit.com/r/rust/comments/108z8pl/how_to_have_a_unique_type/

But basically we could use the above approaches to create a Dim that can be a ConstDim because if you create another version of it, it is unique. And any clones would be the same type & have the same data.

I didn't look too much into it, but it could be a viable alternative to this and give us all the same benefits.

Narsil:
>  give us all the same benefits.

You mean so that `Dyn<'B'>(3)` wouldn't be `Dyn<'B'>(2)` but always `Dyn<'B'>(3)` ?  I think that's already the case (technically there not the same thing, but they resolve to being equal). 

For the lifetime that's interesting, like you wouldn't be able to modify the dynamic shape before dropping all references to it (so the relevant tensors) and so you could know that at any given time in the program the shape would have to be the same.

Maybe you're saying you need both elements. The uniqueness to make sure you can't avoid the compile time thing by just creating a new dimension. 

Is that it ?

coreylowman:
So specifically, it would be able to create a type that is guarunteed to be unique each time you created it. Notably even dimensions that have the same value wouldn't be the same time:
```rust
let a = dim!(3);
// `a` and `b` are different types
let b = dim!(3);

// we can clone `a` to get a copy of it - this will be the same type
let c = a.clone();
```

Which can then be used as a `ConstDim`, which is any type that at compile time has exactly the same value. Each type this macro creates will only ever have a single value.

Now I have no idea about the usability of these things, or if you can even write the type down, but it's an interesting direction as far as I understand this concept.

Cause yeah even with the string names there's nothing stopping you from just creating ones with the same name but different values. If we could have some way to ensure you can't create ones with the same name in multiple places, that wouldn't be an issue (and this is what the lifetime approach does).

jafioti:
Interesting, could this `dim!(n)` be filled in at runtime? I'd think this can be used to create the runtime dimensions, but then in the models, you can specify dynamic dimensions as `D: Unique + Dim` or something. That way we can use the dyn dimensions in a similar way const dims are used,  with `ConstEq` and stuff.

coreylowman:
Going to close this for now - since most operations now support runtime dimensions I'm not sure how needed this is. Definitely willing to revisit in the future

PR #417, opened 2023-01-30
Enable multi-core matmul

### LINKED ISSUE #401: Add threading feature for matrixmultiply when device is CPU
`matrixmultiply` already supports 1 to 4 threads, so its threading feature can possibly be easily integrated to `dfdx`.

I have already manually cargo added ```matrixmultiply = {version = "0.3.2", features = ["threading"]} ``` to the .toml file (right after `dfdx`), but training still looks single-threaded. Needs further investigation.

**Edit**: I was wrong, just activating threading as above does improve performance. I was right that cpu utilization still looks rather single-threaded, but I guess there is a little more utilization of the remaining threads that helps overall performance.

For comparison (on the 06-mnist example with `BATCH_SIZE = 1024` on a Ryzen 5800x3d):
- with `$Env:MATMUL_NUM_THREADS=1; cargo run --release --example 06-mnist -- .\tmp\`:

```
Found 60000 training images
Epoch 0 in 2.718831s (21.333 batches/s): avg sample loss 0.00090
```

- with `$Env:MATMUL_NUM_THREADS=4; cargo run --release --example 06-mnist -- .\tmp\`:

```
Found 60000 training images
Epoch 0 in 1.2467984s (46.519 batches/s): avg sample loss 0.00090
```

**Edit2**: Batch size plays a big role to the performance gains, which makes sense, since the larger the batches, the more the overall load lies on the matmul I guess. Which was also why I initially thought that threading doesn't have an effect.

With the original `BATCH_SIZE = 32` of the example:

- with `$Env:MATMUL_NUM_THREADS=1; cargo run --release --example 06-mnist -- .\tmp\`:

```
Found 60000 training images
Epoch 0 in 5.2134919s (359.644 batches/s): avg sample loss 0.00770
```

- with `$Env:MATMUL_NUM_THREADS=4; cargo run --release --example 06-mnist -- .\tmp\`:

```
Found 60000 training images
Epoch 0 in 4.372751s (428.792 batches/s): avg sample loss 0.00770
```

infalmo:
Closes #401

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ features = ["nightly", "numpy"]
 no-std-compat = { version = "0.4.1", default-features = false, features = [ "alloc", "compat_hash" ] }
 rand = { version = "0.8.5", default-features = false, features = ["std_rng"] }
 rand_distr = { version = "0.4.3", default-features = false, features = ["std_math"] }
-matrixmultiply = { version = "0.3.2", default-features = false }
+matrixmultiply = { version = "0.3.2", default-features = false, features = ["threading"] }
 zip = { version = "0.6.2", default-features = false, optional = true }
 cblas-sys = { version = "0.1.4", default-features = false, optional = true }
 libc = { version = "0.2", default-features = false, optional = true }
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -27,7 +27,7 @@ features = ["nightly", "numpy"]
 no-std-compat = { version = "0.4.1", default-features = false, features = [ "alloc", "compat_hash" ] }
 rand = { version = "0.8.5", default-features = false, features = ["std_rng"] }
 rand_distr = { version = "0.4.3", default-features = false, features = ["std_math"] }
-matrixmultiply = { version = "0.3.2", default-features = false }
+matrixmultiply = { version = "0.3.2", default-features = false, features = ["threading"] }
```
coreylowman, line 30: Since threading feature requires std, let's add this to the std feature in the section below instead (see example I posted in issue)

coreylowman (approved):
Nice!

Issue #156, opened 2022-08-21
Datasets/iterator mega issue


coreylowman:
dfdx should have a set of easy to use datasets built in (with an optional feature). Things like:
- mnist
- cifar10
- coco (maybe?)
- imagenet/imagenette
- some language datasets

This also encompasses things like DataLoader & dataset iterators. It's still an open question about whether a dataloader is needed, or if that would be encompassed by some BatchIterator<T> class.

coreylowman:
Whatever the dataset iterator api is, it should also enable things like torchvision.transforms. E.g. I should be able to do something like:

```rust
let transforms = (CenterCrop(10, 10), Normalize(...));
for x, lbl in dataset.iter_batches::<64>(&mut rng).map(transforms) {
    ...
}
```

coreylowman:
Needs to account for things like the RedCaps dataset which only contains URLs for images, and large datasets which many files and indices

quietlychris:
First, just want to say that I'm really excited to see this crate coming along! It's really cool to start to start seeing convolutional networks, in particular, becoming available in Rust. Second, obviously no pressure to do so, but I wanted to make you aware that that I maintain an existing library for downloading and parsing the CIFAR-10 dataset into Vec<u8> or `ndarray` structs (including options for different version for compatibility) that you might find useful. I'm also open adding new features that you need something that's not available! https://github.com/quietlychris/cifar-ten (also on crates.io)

coreylowman:
Nice work @quietlychris, looks very useful! Will let you know if there's features that would be useful. Definitely need to think through a `trait Dataset` to make writing training loops that are generic over kinds of datasets easier for library devs. Could easily see impl'ing that trait for external datasets like your cifar-ten library

cBournhonesque:
I'll try to work on this; probably will start with an API that's very similar to pytorch

coreylowman:
A sketch of what the frontend UX might look like:

```rust
for (img, label) in dataset.iter_shuffled(&dev).batches(10, false) {
    // img: Tensor<(Dyn, ...), ...>
    // lbl: Tensor<(Dyn, ...), ...>
    ...
}
```

```rust
for (img, label) in dataset.iter(&dev).const_batches::<10>() {
    // img: Tensor<(Const<10>, ...), ...>
    // lbl: Tensor<(Const<10>, ...), ...>
    ...
}
```

[...]

PR #426, opened 2023-01-30
Add download_and_extract() for MNIST dataset with two source options

quietlychris:
This should help make it a little easier for people to try out the MNIST example without needing to manually download and extract the files themselves, although it does slightly increase the number of dev dependencies. I also suggested using the `--release` flag since that speeds up the initial data importing. 

Two options for the download source are available, the standard one from Yann LeCun's site as well as a mirror that I host on my [personal site](https://cmoran.xyz/data) along with the Fashion MNIST, CIFAR-10, and a few misc. ones that the Rust-ML `linfa` library uses. If you'd prefer for me to remove the second option, just let me know.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,7 +46,7 @@ test-cuda = ["cuda"]
 [dev-dependencies]
 rand = "0.8.5"
 tempfile = "3.3.0"
-mnist = "0.5.0"
+mnist = {version = "0.5.0", features = ["download"]}
 indicatif = "0.16.2"
 
 [build-dependencies]
--- a/examples/06-mnist.rs
+++ b/examples/06-mnist.rs
@@ -2,20 +2,13 @@
 //! to build a neural network that learns to recognize
 //! the MNIST digits.
 //!
-//! To download the MNIST dataset, do the following:
-//! ```
-//! mkdir tmp/ && cd tmp/
-//! curl -O http://yann.lecun.com/exdb/mnist/train-images-idx3-ubyte.gz \
-//!     -O http://yann.lecun.com/exdb/mnist/train-labels-idx1-ubyte.gz \
-//!     -O http://yann.lecun.com/exdb/mnist/t10k-images-idx3-ubyte.gz \
-//!     -O http://yann.lecun.com/exdb/mnist/t10k-labels-idx1-ubyte.gz
-//! gunzip t*-ubyte.gz
-//! cd -
-//! ```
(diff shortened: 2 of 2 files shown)


examples/06-mnist.rs
```diff
@@ -36,7 +29,12 @@ struct MnistDataset {
 
 impl MnistDataset {
     fn train(path: &str) -> Self {
-        let mnist: Mnist = MnistBuilder::new().base_path(path).finalize();
+        let mnist: Mnist = MnistBuilder::new()
+            .base_url("http://yann.lecun.com/exdb/mnist/")
+            // .base_url("https://cmoran.xyz/data/mnist/standard/")
```
coreylowman, line 34: Yeah let's just keep the yann lecun ones since that is pretty standard. Great addition though!

Cargo.toml
```diff
@@ -46,7 +46,7 @@ test-cuda = ["cuda"]
 [dev-dependencies]
 rand = "0.8.5"
 tempfile = "3.3.0"
-mnist = "0.5.0"
+mnist = {version = "0.5.0", features = ["download"]}
```
coreylowman, line 49: Also yeah - the github workflow errors were why I didn't add this feature to begin with, since it doesn't work on all OS. Likely will be adding mnist & other datasets directly to dfdx in the future with #156 so we don't need to worry about that.

In the mean time, if you want to contribute the fix to the mnist crate, that'd be great, but it looks like it hasn't been updated in a while on crates.io

nkoppel: I think that this add should actually work on windows, and that the github action fails because it tries to build for the wrong target. If you look at the [file the compile error occurs in] (https://github.com/davidMcneil/mnist/blob/master/src/download.rs) it should have support for windows.

coreylowman: I think that may be due to the fix not being re-released on crates.io? The last comment on the open issue about it asks about a new release. It seems like the 0.5.0 release was nov 8th, 2021, and the fix was merged in 2022

quietlychris: Yes, I think that's the case. I just sent a message to the maintainer to see if they might be willing to publish that patch; it might take a bit for that to come through, however, so I'll keep an eye on it and circle back when there's an update

coreylowman: Sounds good, thanks for reaching out

coreylowman:
Going to close this for now. Let's revisit when they push a new version.

PR #431, opened 2023-02-04
Possibly fix `atomicMaxf` and `atomicMinf` for doubles

ViliamVadocz:
I implemented my suggestion of converting to `long long int` in the atomic max/min functions. I could not test it since the Rust portion of the code does not compile.

I also added a missing header that is needed to use `__half`.


coreylowman:
Thanks!

PR #433, opened 2023-02-06
Adding nn builder structs, dtype generics, and remove device defaults.

### LINKED ISSUE #419: Remove Device default value for layers in `src/nn`
See discussion in #410 for reasoning for this.

Basically this diff for all the layers that have a device generic:

```diff
-pub struct Linear<const I: usize, const O: usize, D: Device<f32> = Cpu> {
+pub struct Linear<const I: usize, const O: usize, D: Device<f32>> {
```

For ergonomics, something may be added in future, but that will be a separate issue.

coreylowman:
Resolves #419 

In order to support multiple dtypes, the nn api needs to have a generic over dtype, similar to tensors. Additionally, the device generic being defaulted to Cpu was causing issues with other APIs, so this PR removes it.

The above two changes meant that any specification of a nn module required also specifying the dtype & device, which could quickly grow out of hand.

This PR thus introduces a number of "builder" unit structs that implement the `BuildOnDevice` trait, so you don't have to specify either dtype or device.

Here is Conv2D as an example:
```rust
pub mod builder {
    #[derive(Debug)]
    pub struct Conv2D<
        const IN_CHAN: usize,
        const OUT_CHAN: usize,
        const KERNEL_SIZE: usize,
        const STRIDE: usize = 1,
        const PADDING: usize = 0,
    >;
}

impl<const I: usize, const O: usize, const K: usize, const S: usize, const P: usize, D>
    BuildOnDevice<D, f32> for builder::Conv2D<I, O, K, S, P>
where
    D: Device<f32>,
{
    type Built = Conv2D<I, O, K, S, P, f32, D>;
    fn try_build_on_device(device: &D) -> Result<Self::Built, <D>::Err> {
        Self::Built::try_build(device)
    }
}
```

Here we see that we can get default generics for free on builder structs, and utilize BuildOnDevice to build the actual struct.

# Considerations for future
- we now have `nn::conv::builder::Conv2D` and `nn::conv::Conv2D`, which have the same name. this could cause confusion
- nn now exposes two sub-modules: `modules` and `builders`. The prior should be used when building a new network layer, and the second should be used when declaring a network structure

### DIFF
--- a/src/nn/linear.rs
+++ b/src/nn/linear.rs
@@ -162,19 +162,13 @@ mod tests {
     ];
     const B: [f32; 2] = [0.3765365, -0.290717];
 
-    #[cfg(feature = "cuda")]
     #[test]
     fn test_linear_ondevice() {
-        use super::super::module::OnDevice;
-
-        let cuda: Cuda = Default::default();
-        let _: Linear<1, 1, _> = BuildModule::build(&cuda);
-        let _: OnDevice<Linear<1, 1>, Cuda> = BuildModule::build(&cuda);
-        let _: OnDevice<(Linear<1, 2>, Linear<2, 1>), Cuda> = BuildModule::build(&cuda);
-
-        let _: Linear<1, 1, Cuda> = Linear::<1, 1>::build_on_device(&cuda);
-        let _: Linear<1, 1, _> = Linear::<1, 1>::build_on_device(&cuda);
-        let _ = Linear::<1, 1>::build_on_device(&cuda);
+        let dev: TestDevice = Default::default();
+        let _: Linear<1, 1, _, _> = BuildModule::build(&dev);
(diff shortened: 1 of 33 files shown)


src/nn/linear.rs
```diff
@@ -136,7 +154,6 @@ impl<'a, B: Dim, S: Dim, const M: usize, D: Device<f32>, T: Tape<D>>
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::nn::BuildOnDevice;
     use crate::{nn::tests::SimpleUpdater, tests::*, unique_id::HasUniqueId};
 
     const W: [[f32; 5]; 2] = [
```
coreylowman, line 159: note to self: test on cuda - the on device stuff will likely have to change

PR #437, opened 2023-02-08
Start of a continuous rl ppo example

usbalbin:
The beginning of an attempt to make a continuous rl ppo example.

I am trying to roughly follow what is done in https://github.com/seungeunrho/minimalRL/blob/master/ppo-continuous.py and the rest is based on examples/rl-ppo.rs. Not at all sure of what I am doing, I will likely have tons of questions... :)

### DIFF
--- a/examples/rl-ppo-continuous.rs
+++ b/examples/rl-ppo-continuous.rs
@@ -8,10 +8,14 @@ use dfdx::{
         BuildModule, Module,
     },
     optim::{Adam, AdamConfig},
-    prelude::{smooth_l1_loss, GradientUpdate, ParamUpdater, UnusedTensors},
-    tensor::{AsArray, Cpu, HasErr, SplitTape, Tensor1D, Tensor2D},
-    tensor_ops::MeanTo,
+    prelude::{smooth_l1_loss, GradientUpdate, ParamUpdater, UnusedTensors, Optimizer, OwnedTape},
+    shapes::Shape,
+    tensor::{
+        AsArray, Cpu, HasErr, SampleTensor, SplitTape, Tensor, Tensor1D, Tensor2D, TensorFrom, Tensor0D,
+    },
+    tensor_ops::{MeanTo, TryDiv, Backward},
 };
+use std::f32;
 use std::time::Instant;
 
 const BATCH_SIZE: usize = 64;
@@ -63,12 +67,12 @@ impl<const IN: usize, const INNER: usize, const OUT: usize> GradientUpdate<Cpu, 
 }
 
 // impl Module for single item
-impl<const IN: usize, const INNER: usize, const OUT: usize> nn::Module<Tensor1D<IN>>
+impl<const IN: usize, const INNER: usize, const OUT: usize, T: Tape<Cpu>> nn::Module<Tensor1D<IN, T>>
     for Network<IN, INNER, OUT>
 {
-    type Output = (Tensor1D<OUT>, Tensor1D<OUT>, Tensor1D<OUT>);
+    type Output = (Tensor1D<OUT, T>, Tensor1D<OUT, T>, Tensor1D<OUT, T>);
 
-    fn forward(&self, x: Tensor1D<IN>) -> Self::Output {
+    fn forward(&self, x: Tensor1D<IN, T>) -> Self::Output {
         let x = self.l1.forward(x);
         (
             self.mu.forward(x),
@@ -125,41 +129,65 @@ fn main() {
         },
     );
 
-    let mut state: Tensor1D<STATE_SIZE> = init_state();
+    let mut state = init_state();
 
     // run through training data
     for _i_epoch in 0..15 {
         let start = Instant::now();
 
-        // <calc advantage>
-        let (new_state, reward, done) = step_simulation(action);
+        // <>
+        let (old_log_prob, action) = {
+            // TODO: Should we avoid calculating _v?
+            let (mu, std, _v)= net.forward(state);
+            let action: Tensor1D<ACTION_SIZE> = dev.sample_normal() * std + mu;
+            let log_prob = {
+                let variance = std.powi(2);
+                -(action - mu).powi(2) / (variance * 2.0)
+                    - std.ln()
+                    - (2.0f32 * f32::consts::PI).sqrt().ln()
+            };
+            (log_prob /*(std, mu, action)*/, action)
+        };
 
-        let td_target = reward + net.value.forward(net.l1.forward(new_state)) * done * GAMMA;
-        let delta = td_target - net.value.forward(net.l1.forward(state));
+        let (new_state, reward, done) = step_simulation(state, &action);
+
+        // <calc advantage>
+        let td_target: Tensor1D<ACTION_SIZE> =
+            net.value.forward(net.l1.forward(new_state)) * f32::from(done as u8) * GAMMA + reward;
+        let delta = td_target.trace() - net.value.forward(net.l1.forward(state.trace()));
         let delta = delta.array();
 
         let mut advantage = 0.0;
-        let advantage = delta.take(delta.len() - 1).map(|delta| {
-            advantage = GAMMA * LAMBDA * advantage + delta[0];
-            advantage
-        });
+        let advantage: Tensor1D<ACTION_SIZE, _> = dev.tensor(
+            TryInto::<[_; ACTION_SIZE]>::try_into(delta
+                .into_iter()
+                .take(delta.len() - 1)
+                .map(|delta| {
+                    advantage = GAMMA * LAMBDA * advantage + delta;
+                    advantage
+                })
+                .collect::<Vec<f32>>()
+                .as_slice())
+                .unwrap(),
+        );
         // </calc advantage>
 
-        let (mu, std, v) = net.forward(state);
-        //let dist = mu.array().into_iter().zip(std.array()).map(|(mu, std)| Normal::new(mu, std)).collect(); // TODO: is this correct? How should this be done in a more "Tensor"y way?
-
-        let log_prob: Tensor1D<INNER_SIZE> = todo!();
+        let (mu, std, value): (Tensor1D<ACTION_SIZE, OwnedTape<Cpu>>, Tensor1D<ACTION_SIZE, OwnedTape<Cpu>>, Tensor1D<ACTION_SIZE, OwnedTape<Cpu>>) = net.forward(state.trace());
+        let log_prob: Tensor1D<ACTION_SIZE, OwnedTape<Cpu>> = {
+            let variance = std.powi(2);
+            -(action.trace() - mu).powi(2) / (variance * 2.0) - std.ln() - (2.0f32 * f32::consts::PI).sqrt().ln()
+        }/*log_prob(std, mu, action)*/;
 
         // ratio = P(action | state, pi_net) / P(action | state, target_pi_net)
         // but compute in log space and then do .exp() to bring it back out of log space
         let ratio = (log_prob - old_log_prob).exp();
 
         // because we need to re-use `ratio` a 2nd time, we need to do some tape manipulation here.
-        let surr1 = ratio.with_empty_tape() * advantage.clone();
-        let surr2 = ratio.clamp(1.0 - EPS_CLIP, 1.0 + EPS_CLIP) * advantage.clone();
+        let surr1: Tensor1D<ACTION_SIZE, OwnedTape<Cpu>> = ratio.with_empty_tape() * advantage.clone();
+        let surr2: Tensor1D<ACTION_SIZE, OwnedTape<Cpu>> = ratio.clamp(1.0 - EPS_CLIP, 1.0 + EPS_CLIP) * advantage.clone();
 
-        let loss = -surr2.minimum(surr1) + smooth_l1_loss(v, td_target, 1.0);
-        let loss = loss.mean();
+        let loss = -surr2.minimum(surr1) + smooth_l1_loss(value, td_target, 1.0).array();
+        let loss: Tensor0D<OwnedTape<Cpu>> = loss.mean();
 
         let loss_v = loss.array();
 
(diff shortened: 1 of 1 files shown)


examples/rl-ppo-continuous.rs
```diff
@@ -0,0 +1,191 @@
...
+const ACTION_SIZE: usize = 2;
+
+/// Custom model struct
+#[derive(Clone)]
+struct Network<const IN: usize, const INNER: usize, const OUT: usize> {
+    l1: (Linear<IN, INNER, f32, Cpu>, ReLU),
+    mu: (Linear<INNER, OUT, f32, Cpu>, Tanh),
+    std: (Linear<INNER, OUT, f32, Cpu>, ReLU), // TODO: should this be SoftPlus?
+    value: Linear<INNER, OUT, f32, Cpu>,
+}
```
usbalbin, line 29: From what I can tell, l1 is always used as the first step and its result is the input the the three other functions. Sometimes we want the result from all functions and sometimes only one or some of them.

Is this how such a network should be modelled? I have tried looking at `07-custom-module.rs`. How on earth is the gradient and optimizer stuff supposed to fit in with this?

usbalbin: Might it perhaps be as simple as `dev.sample_normal() * std + mu`?

examples/rl-ppo-continuous.rs
```diff
@@ -0,0 +1,200 @@
...
+
+        // TODO: How to nn.utils.clip_grad_norm_(self.parameters(), 1.0)
+
+        // run backprop
+        let gradients = loss.backward();
+
+        // update weights with optimizer
+        optimizer
+            .update(&mut net, gradients)
+            .expect("Unused params");
```
nkoppel, line 174: "Update" failing does not mean that there were unused parameters, it means that something went wrong allocating memory or executing operations.

examples/rl-ppo-continuous.rs
nkoppel: > Is this how such a network should be modelled? I have tried looking at `07-custom-module.rs`. How on earth is the gradient and optimizer stuff supposed to fit in with this?

This is probably the best way to model this kind of network, at least in terms of being able to use optimizers ergonomically. Your implementation of `GraidentUpdate` seems correct, as it closely mirrors the implementation of `GradientUpdate` for [transformer encoders](https://github.com/coreylowman/dfdx/blob/f9d6aa0ac02102057980e891c85eccf5c1f298dd/src/nn/transformer/encoder.rs#L120).

> Might it perhaps be as simple as `dev.sample_normal() * std + mu`?
>
> Edit: no, dist is not the supposed to be the Tensor, it is the normal distribution generator thing right? [Here](https://github.com/seungeunrho/minimalRL/blob/master/ppo-continuous.py#L138) it is used to generate the next action (a tensor).

Dfdx does not currently have an object that represents a random distribution, so you should just use `dev.sample_normal() * std + mu` in place of `dist.sample()`. This should be valid, since [`Normal`](https://pytorch.org/docs/stable/distributions.html#normal) uses `mu` as `loc` and `std` as `scale`.

usbalbin: Ah, Thanks a lot.

How would I accomplish something like [this](https://github.com/seungeunrho/minimalRL/blob/master/ppo-continuous.py#L139):

```python
dist = Normal(mu, std)
a = dist.sample()
log_prob = dist.log_prob(a) # <-- This
```

and [this](https://github.com/seungeunrho/minimalRL/blob/master/ppo-continuous.py#L139)? What does [it](https://pytorch.org/docs/stable/distributions.html#torch.distributions.normal.Normal.log_prob) mean? I have read [this](https://stackoverflow.com/questions/54635355/what-does-log-prob-do) several times now, can't say I got that much wiser... :) Something about the logarithm of the probability for the value of action in the distribution? Does this mean it is not something we can do on our sample (dist), but rather have to calculate from action, `mu`, `std` and how the normal distribution behave?

nkoppel: From what I understand, `log_prob` computes `log(pdf(x))`, where pdf is the [normal distribution's probability distribution function](https://en.wikipedia.org/wiki/Normal_distribution). Dfdx doesn't currently have an implementation of pdfs/cdfs of probability distributions, but it should be pretty easy to implement directly, and you can reference [pytorch's implementation](https://pytorch.org/docs/stable/_modules/torch/distributions/normal.html#Normal.log_prob).

usbalbin: Again, thanks a lot :)

examples/rl-ppo-continuous.rs
```diff
@@ -0,0 +1,241 @@
...
+    for Network<IN, INNER, OUT>
+{
+    type Output = (Tensor1D<OUT, T>, Tensor1D<OUT, T>, Tensor1D<OUT, T>);
+
+    fn forward(&self, x: Tensor1D<IN, T>) -> Self::Output {
+        let x = self.l1.forward(x);
+        (
+            self.mu.forward(x),
+            self.std.forward(x),
+            self.value.forward(x),
```
usbalbin, line 80: How do clone `x`? `OwnedTape<Cpu>` does not implement `Copy` and I can not pass x by reference. I am not sure I understand the concept of Tape. Will try to google it another day...

nkoppel: I believe that the way to do this is to do `x.retaped()` to clone the data of a tensor without cloning it's `Tape`. Since tapes serve to keep a record of what backward operations need to be done once `backward` is called, you will need to be sure that the output using the non-retaped x is always used to calculate the loss. Otherwise, the tape will not contain the operations done by `l1`, and `l1` will not be trained.

nkoppel: I've actually decided that I like that tapes cannot be cloned from a design standpoint, because it prevents users from calculating gradients inefficiently, i.e. with multiple calls to backwards instead of only one. If you use `retaped` in Network's `forward`, you shouldn't encounter any issues, since you end up using all recombining all of it's outputs when calculating the gradient.

examples/rl-ppo-continuous.rs
```diff
@@ -0,0 +1,272 @@
...
+
+        // TODO: How to nn.utils.clip_grad_norm_(self.parameters(), 1.0)
+
+        // run backprop
+        let gradients = loss.backward();
+
+        // update weights with optimizer
+        optimizer
+            .update(&mut net, gradients)
+            .expect("Failed to update");
```
usbalbin, line 203: I assume I'm doing something wrong. Still don't quite understand when and how to handle the tape thing. Correct me if I am wrong but conceptually the tape is like some sort of history so that we can trace the loss backwards all the way to some sort of beginning. How this applies in this example is still not clear though. I have tried to look at examples/rl-ppo and the [python example](https://github.com/seungeunrho/minimalRL/blob/master/ppo-continuous.py) to try to guesstimate the tapes things but still...

```
thread 'main' panicked at 'Failed to update: UnusedParams(UnusedTensors { ids: [UniqueId(0), UniqueId(1), UniqueId(6), UniqueId(7)] })', examples/rl-ppo-continuous.rs:203:14
```

examples/rl-ppo-continuous.rs
```diff
@@ -0,0 +1,272 @@
...
+    let mut state = init_state(&dev);
+
+    // run through training data
+    for _i_epoch in 0..15 {
+        let start = Instant::now();
+
+        // <>
+        let (old_log_prob, action) = {
+            // TODO: Should we avoid calculating _v? _v is the one with the original tape if that matters?
+            let (mu, std, _v) = net.forward(state.retaped());
```
usbalbin, line 134: In our implementation of nn::Module, `value` is the only one of three that is not a `retaped`. Does this mean that `value` is the only one that has the complete history including `l1`? Is this then (one of) the reason for the below error about unused parameters?

usbalbin: Changing to:

```rust
fn forward(&self, x: Tensor2D<BATCH, IN, T>) -> Self::Output {
    let x = self.l1.forward(x);

    let std = self.std.forward(x.retaped());
    let value = self.value.forward(x.retaped());
    (
        self.mu.forward(x),
        std,
        value
    )
}
```

Makes `UniqueId(0), UniqueId(1)` go away from the error below

usbalbin:
```
let loss = -surr2.minimum(surr1) + smooth_l1_loss(value, td_target, 1.0).array();
```
Using .array() to go through a plain f32 nukes the tape thing. Now using broadcast instead... Seems to resolve that error

usbalbin:
I am not really seing any improvement even after thousands of iterations. Am I doing something wrong? Was skipping the minibatch thing an invalid simplification? Any ideas? :)

examples/rl-ppo-continuous.rs
```diff
@@ -0,0 +1,274 @@
...
+const ACTION_SIZE: usize = 1;
+
+/// Custom model struct
+#[derive(Clone, Debug)]
+struct Network<const IN: usize, const INNER: usize, const OUT: usize> {
+    l1: (Linear<IN, INNER, f32, Cpu>, ReLU),
+    mu: (Linear<INNER, OUT, f32, Cpu>, Tanh),
+    std: (Linear<INNER, OUT, f32, Cpu>, ReLU), // TODO: should this be SoftPlus?
+    value: Linear<INNER, OUT, f32, Cpu>,
+}
```
coreylowman, line 31: SplitInto layer was added just for this purpose 😀 
```rust
type Network<const IN: usize, const INNER: usize, const OUT:usize> = (
    (Linear<IN, INNER, f32, Cpu>, ReLU),
    SplitInto<(
        (Linear<INNER, OUT, f32, Cpu>, Tanh),
        (Linear<INNER, OUT, f32, Cpu>, ReLU),
        Linear<INNER, 1, f32, Cpu>
    )>
);
```

usbalbin: Oh, thanks :)

TimerErTim:
> I am not really seeing any improvement even after thousands of iterations. Am I doing something wrong? Was skipping the minibatch thing an invalid simplification? Any ideas? :)

Don't know if you have the same problem but have you checked the calculated gradients? Somehow after the first training step they are always 0.0 for me. Maybe you have the same problem...

I am currently working on a minimal reproducable example.

usbalbin:
Printing the nonzero gradients of the last iterations gives this:

```
State: temp=-40.03252, power=-0.2873001
net.0.0.bias: StridedArray { data: [0.0, 0.0, 0.52112234], shape: (Const,), strides: [1] }
net.0.0.weight: StridedArray { data: [0.0, 0.0, 0.52112234], shape: (Const,), strides: [1] }
net.1.0.2.0.bias: StridedArray { data: [1.0], shape: (Const,), strides: [1] }
net.1.0.2.0.weight: StridedArray { data: [0.0, 0.0, 0.5072972], shape: (Const, Const), strides: [3, 1] }
loss=80.84794 in 1.01211ms
State: temp=-40.31982, power=-0.0022200346
net.0.0.bias: StridedArray { data: [0.0, 0.0, 0.5209628], shape: (Const,), strides: [1] }
net.0.0.weight: StridedArray { data: [0.0, 0.0, 0.5209628], shape: (Const,), strides: [1] }
net.1.0.2.0.bias: StridedArray { data: [1.0], shape: (Const,), strides: [1] }
net.1.0.2.0.weight: StridedArray { data: [0.0, 0.0, 0.18710533], shape: (Const, Const), strides: [3, 1] }
loss=81.08815 in 972.044µs
State: temp=-40.32204, power=-0.6552692
net.0.0.bias: StridedArray { data: [0.0, 0.02987676, 0.0], shape: (Const,), strides: [1] }
net.0.0.weight: StridedArray { data: [0.0, 0.02987676, 0.0], shape: (Const,), strides: [1] }
net.1.0.2.0.bias: StridedArray { data: [1.0], shape: (Const,), strides: [1] }
net.1.0.2.0.weight: StridedArray { data: [0.0, 0.0011303872, 0.0], shape: (Const, Const), strides: [3, 1] }
loss=80.8971 in 988.113µs
```

suggesting that the gradients of `mu` (net.1.0.0) and `std`(net.1.0.1) are always zero

TimerErTim:
That is also what I observed in #486 

The problem seemed to stem from using clamping activation functions in the output layers. In your current branch I see you use ReLU for std and Tanh for mu. That would explain why both lose meaningful gradients, though I am not sure why Tanh would cause such a problem. Maybe you can try removing those activation functions.

coreylowman:
Has anyone checked if the same 0.0 gradients thing occurs in pytorch with this setup as well?

TimerErTim:
Honestly I haven't but this seems to be a universally observable phenomenon called "dead neurons". The model learns a bias that always results in a 0 output after the last ReLU and since 0 is a constant and the derivative of a constant is 0, the gradient is also 0. This prevents the bias from changing however effectively deadlocking the whole network's learning process.

Reasons for such a learned bias may be almost always 0.0 target values during training. That also aligns with my project since the network rarely gets a different target sample than simply zeros.

coreylowman:
@usbalbin what's the status of this?

usbalbin:
Have not had time to work on this for a while. Last thing I got stuck on was that I could not quite seem to get the network to improve.

I do not know when/if I get time. Feel free to close in the meantime, or if anyone else want to pick up where I left of.

nkoppel:
Have you tried running it after #507/#510? This bug may have been causing some of the gradients to be zero.

coreylowman:
Closing because stale. Feel free to reopen if you start working on it again

examples/rl-ppo-continuous.rs
```diff
@@ -0,0 +1,206 @@
...
+
+        // <>
+        let (old_log_prob, action) = {
+            // TODO: Should we avoid calculating _v? _v is the one with the original tape if that matters?
+            let (mu, std, _v) = net.forward(state.retaped::<OwnedTape<Cpu>>());
+
+            let action: Tensor1D<ACTION_SIZE> = dev.sample_normal() * std.clone() + mu.clone();
+            let log_prob = {
+                let variance = std.clone().powi(2);
+                -(action.clone() - mu).powi(2) / (variance * 2.0)
```
mileswatson, line 67: I think variance should be squared?

PR #442, opened 2023-02-09
Add synchronize method to Cuda device

### LINKED ISSUE #383: Add synchronize method to Cuda device
This will be useful when benchmarking to ensure that the kernels have finished executing

ViliamVadocz:
Closes #383 

I am not sure how to properly test that this works. It doesn't error but I did not observe any difference between using it and not using it. Maybe all the kernels I have been trying are too small?


coreylowman (approved):
This looks good to me - yeah probably wouldn't notice unless the kernel is big enough (e.g. the whole forward pass of `examples/nightly-resnet18.rs`)

PR #443, opened 2023-02-11
Generalized Embeddings and added LearnedPositionalEmbedding

jafioti:
The embedding BATCH and SEQ dims were needlessly Const, so I changed them to `Dim` which will allow for dynamic dimensions. 

I also added a first pass implementation of LearnedPositionalEmbedding which is needed for transformers.


jafioti:
I also made a small change (4 lines) to add a second residual connection in the encoder and a third in the decoder. This is how the attention is all you need transformer works.

src/nn/embedding.rs
```diff
@@ -129,6 +135,134 @@ impl<const VOCAB: usize, const DIM: usize, D1: Device<f32>, D2: Device<f32>> ToD
...
+    for LearnedPositionalEmbedding<MAX_LEN, DIM, f32, D>
+where
+    D: TensorFrom<std::vec::Vec<usize>, (SEQ,), usize>,
+{
+    type Output = Tensor<(SEQ, Const<DIM>), f32, D, T>;
+    fn forward(&self, input: Tensor<(SEQ, Const<DIM>), f32, D, T>) -> Self::Output {
+        let (input, tape) = input.split_tape();
+        let positions: Tensor<(SEQ,), usize, D> = self
+            .device
+            .tensor((0..input.shape().0.size()).collect::<std::vec::Vec<_>>());
```
coreylowman, line 187: Nice 😀 Do you need the vec generic parameter to collect? Would've thought rust could infer that

src/nn/embedding.rs
```diff
@@ -129,6 +135,134 @@ impl<const VOCAB: usize, const DIM: usize, D1: Device<f32>, D2: Device<f32>> ToD
...
+/// let mut model = Model::build_on_device(&dev);
+/// // single sequence of ids
+/// let inputs: Tensor<Rank1<5>, usize, _> = dev.zeros();
+/// let _: Tensor<(Const<5>, Const<2>,), f32, _> = model.forward(inputs);
+/// // batched sequence of ids
+/// let inputs: Tensor<Rank2<10, 5>, usize, _> = dev.zeros();
+/// let _: Tensor<(Const<10>, Const<5>, Const<2>), f32, _> = model.forward(inputs);
+/// ```
+#[derive(Debug, Clone)]
+pub struct LearnedPositionalEmbedding<
```
coreylowman, line 164: Let's move this to it's own file, also thoughts on just calling it `PositionalEmbedding`?

src/nn/transformer/encoder.rs
```diff
@@ -161,8 +161,10 @@ where
         let x = self.self_attn.forward(src.clone().put_tape(tape));
         let x = x + src;
         let x = self.norm1.forward(x);
-        let x = self.ff.forward(x);
-        self.norm2.forward(x)
+        let (x, tape) = x.split_tape();
+        let x_residual = x.clone();
+        let x = self.ff.forward(x.put_tape(tape));
+        self.norm2.forward(x + x_residual)
```
coreylowman, line 167: Let's split this out into a separate PR to discuss - is this what pytorch does? I'm surprised I was able to get the same numbers as pytorch without having this if so

coreylowman (changes requested):
Nice contribution, just some reorganization requests before merging 👍

src/nn/embedding.rs
Narsil: I would advise actually against adding such a layer. 

It's only extremely thin addition on top of `Embedding`. It's not very flexible (for instance roberta position ids start at 2 for some reason https://github.com/huggingface/transformers/issues/10736 which this has no way to represent). 

Seems like this belongs in user code imho.

Issue #425, opened 2023-01-30
Add EMA (exponential moving average) support for NN


coreylowman:
This would let you take two models and keep a moving average of them over time.

Potential interfaces would be:
1. `impl Add/Mul<f32>` for layers so you can do `ema = ema * 0.9 + model * 0.1`
2. Add a separate struct to handle this `struct EMA<M>`

nkoppel:
I think it would be even better if we implemented the assignment operators (`+=`, `*=`, etc), for tensors and layers. Then, ema could be implemented as:
```rust
ema *= 1.0 / a -  1.0
ema += model
ema /= 1.0 / a
```
where a is the weight of `model` in the weighted average.

This implementation will save a substantial amount of vram for large models, and would prevent extra allocations.

coreylowman:
Should be very easy to implement this once we can perform ema for tensors with the new visitors api

coreylowman:
Note for future: use the new tensor ema method in BatchNorm2d:

```rust
self.running_mean = self.running_mean.clone() * (E::ONE - self.momentum)
            + mean_chan.retaped::<NoneTape>() * self.momentum;
self.running_var = self.running_var.clone() * (E::ONE - self.momentum)
            + var_chan.retaped::<NoneTape>() * (self.momentum * n / (n - E::ONE));
```

could be

```rust
self.running_mean.ema(mean_chan.retaped::<NoneTape>(), self.momentum);
self.running_var.ema(var_chan.retaped::<NoneTape>() * (n / (n - E::ONE)), self.momentum);
```

Issue #400, opened 2023-01-25
Gradient Accumulation


jafioti:
Often times it is desired to train on larger batch sizes than can fit on the GPU / memory at once. Accumulating gradients across mini-batches is a solution, which can effectively simulate larger batch sizes, albeit without the parallelism advantage.

A straightforward way to do this would be to impl `Add<Gradients<D>>` to `Gradients<D>` such that gradients on the same device can be added.

I'm not sure how this would be handled, since the grads seem to be stored in a Box<dyn Any>, which I don't see how you can add to.

coreylowman:
I think if we change the underlying storage to `Box<dyn std::ops::AddAssign>`, we could do this. Probably prefer add assign so we aren't cloning things.

Would need to add implementations of AddAssign for raw device storage

coreylowman:
So I was working on this a bit cause I thought I had a pretty clever solution. Turns out `Box<dyn Any>` makes this pretty hard due to object safety rules.

Notably you can't have a trait that takes/returns any `Self` objects. So you can't add `AddAssign` or something like:

```rust
trait AddSelf {
    fn add_self(self, rhs: Self) -> Self;
}
```

If we can't do this with Gradients object, we may need to add some separate gradient accumulator object that does lazy addition:

```rust
struct GradientAccumulator {
    gradients: Vec<Gradients>
}
```

and then add some abstraction layer for the optimizers to use:

```rust
trait HasGradients {
    fn remove<T>(&mut self, t: &T) -> Option<T::Gradient>
    where
        T: HasUniqueId + AllocGrad;
}
```

impl this for both `Gradients` and `GradientAccumulator`, and then change optimizer to accept this trait:

```rust
fn update<G: HasGradients>(
        &mut self,
        module: &mut M,
        gradients: G,
    ) -> Result<(), OptimizerUpdateError<D>>;
```

jafioti:
@coreylowman Why can't there be a trait that accepts / returns Self? That first trait you had looks acceptable to me

coreylowman:
You can't use that kind of object with `dyn`. I.e. you can't have `Box<dyn AddSelf>` because the AddSelf trait is not object safe. See https://doc.rust-lang.org/reference/items/traits.html#object-safety

coreylowman:
Another option for this: give an option to pass in an *existing* Gradients object to `.trace()`.

PR #444, opened 2023-02-12
[WIP] Factor out nn impls using TupleLists and the GetTensors trait.

### LINKED ISSUE #435: Parameter Count
```python
```

```rust
pub trait CountParams {
}
```
```rust
impl <
    S: Shape,
    E: Dtype + Unit,
    T,
        self.shape().num_elements()
    }
}
```

```rust
trait ToStateDict {
}
```

nkoppel:
A draft of a solution to the issues raised on #435 using TupleLists, which are nested tuples of the form `(T1, (T2, (T3, ())))`. This data structure is fairly simple to serialize modules into and to perform operations on without sacrificing too much performance or sacrificing type-checking at compile time. TupleLists can also be zipped, much like iterators, allowing operations between multiple Modules of the same type to be implemented.

While this will definitely be good for internal code, I am not super happy how GetTensors exposes users who implement custom modules to TupleLists, as specifying the type of a TupleList as the result of appending other TupleLists is pretty unwieldy right now, with `<<Linear<2, 3, f32, TestDevice> as GetTensors>::Tensors as TupleList>::Appended<<Linear<3, 5, f32, TestDevice> as GetTensors>::Tensors>` being the type definition for simply appending two linear layers.

Closes #435, and may make #400 and #425 easier to implement.

Tasks:

- [ ] Implement `GetTensors` for all Modules (`Repeated` may need special support)
- [ ] Implement `GradientUpdate` using `GetTensors`
- [ ] Implement `ResetParams` using `GetTensors` (`linear` and `conv`/`conv2d` will present some issues)
- [ ] Implement `SaveToNpz` using `GetTensors`
- [ ] Implement `LoadFromNpz` using `GetTensors`

### DIFF
--- a/src/nn/mod.rs
+++ b/src/nn/mod.rs
@@ -126,6 +126,7 @@ mod repeated;
 mod residual;
 mod split_into;
 mod transformer;
+mod tuple_lists;
 
 pub use module::*;
 
--- a/src/nn/tuple_lists.rs
+++ b/src/nn/tuple_lists.rs
@@ -1,0 +1,206 @@
+use std::vec::Vec;
+
+use crate::prelude::{DeviceStorage, Dtype, HasErr, HasShape, Shape, Tensor};
+
+pub type TensorNames<A> = <A as GetTensors>::Names;
+pub type Tensors<'a, A> = <A as GetTensors>::Tensors<'a>;
+pub type TensorsMut<'a, A> = <A as GetTensors>::TensorsMut<'a>;
+pub type TupleAppend<A, B> = <A as TupleList>::Appended<B>;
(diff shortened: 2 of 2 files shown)


coreylowman:
Yeah this feels quite complex, and given that there may be special cases needed for certain modules I'm not convinced it will necessarily be good for internal code.

I could see EMA being implemented fairly straightforwardly with this, but this wouldn't help gradient accum unless we move gradients to use this tuple list set up.

nkoppel:
Yeah, I have to agree that this approach is over-complicated in practice. TupleLists also don't really feel like they belong in dfdx, and implementing the GetTensors trait for end users is unwieldy. It has also been very difficult to implement GradientUpdate using TupleFold due to issues with lifetimes.

PR #460, opened 2023-02-16
[WIP] Add VisitTensors traits and simplify nn internals.

### LINKED ISSUE #435: Parameter Count

nkoppel:
Adds framework to run functions on all sets of corresponding tensors in groups of immutable and mutable references to modules of a single type, as discussed in #435. Closes #435, and will make #425 easier to implement.

Tasks:

- [x] Implement VisitTensorGroups for all Modules in `nn`
- [x] Implement GradientUpdate (disable for some fields in batchnorm2d)
- [x] Implement ResetParams (linear and conv2d need special treatment)
- [x] Implement SaveToNpz
- [x] Implement LoadFromNpz
- [ ] Document internals in visit_tensors.rs
- [ ] Implement VisitTensorGroups for Mlp in 07-custom-module


src/nn/visit_tensors.rs
```diff
@@ -0,0 +1,177 @@
...
+
+pub struct ModuleGroup<'a, const N: usize, const M: usize, T> {
+    pub refs: [&'a T; N],
+    pub refs_mut: [&'a mut T; M],
+    pub name: Option<String>,
+}
+
+// TODO? : Prevent heap allocation using unsafe or ArrayVec
+fn map_mut<'a, A: 'a + Debug, B: 'a + Debug, const N: usize, F: FnMut(&'a mut A) -> &'a mut B>(
+    arr: &'a mut [A; N],
```
coreylowman, line 17: If it's possible to make this `[&'a mut A; N]` then you can just call map directly

src/nn/visit_tensors.rs
```diff
@@ -0,0 +1,177 @@
+use crate::shapes::Dtype;
+
+use crate::prelude::*;
+use std::{
+    fmt::Debug,
+    string::{String, ToString},
+};
+
+pub struct ModuleGroup<'a, const N: usize, const M: usize, T> {
```
coreylowman, line 9: What's the reasoning behind this struct? I think each module can just call visit separately on each tensor it holds, which would get rid of the need for this

src/nn/visit_tensors.rs
```diff
@@ -0,0 +1,177 @@
...
+}
+
+pub trait TensorVisitor<const N: usize, const M: usize, E: Dtype, D: DeviceStorage> {
+    type Err: std::fmt::Display + Debug;
+
+    fn call<S: Shape>(
+        &mut self,
+        tensors: ModuleGroup<N, M, Tensor<S, E, D>>,
+    ) -> Result<(), Self::Err>;
+    fn set_option(&mut self, _option: TensorVisitorOption) {}
```
coreylowman, line 74: Probably could just pass the option as a parameter into call?

coreylowman (commented):
What's your sense for if this will actually help internals? If we could figure out how to use this to implement EMA, I think that'd be a key advantage, but I haven't been able to figure out how to do "zipped" visiting

src/nn/visit_tensors.rs
nkoppel: This is for "zipped" visiting, it holds a group of N immutable and M mutable references to a module, and has a convenience method for creating a new ModuleGroup that holds references to a field of each referenced Module.

src/nn/visit_tensors.rs
nkoppel: This is necessary because `array.map` consumes the original array. We are able to use `array.map` for immutable references because they implement Copy, but mutable references don't implement Copy or Clone.

src/nn/visit_tensors.rs
nkoppel: I've gone ahead and implemented this suggestion, because lets us reduce the amount of state we have to keep track of in TensorVisitors.

nkoppel:
I think that this pr will grant us a lot of flexibility in implementing new modules and new functionality for modules. It will also make it much easier for users who implement custom modules to get access to a lot of features that they otherwise would've had to write a lot of boilerplate to use.

Also, I don't think that implementating VisitTensorGroups is very complicated, because it pretty much amounts to specifying how to access each field, and defining each field's name. TensorVisitor is also pretty simple to implement, as it amounts to getting the tensors you need from `tensors` and passing them to a function. For example, implementing CountParams for every nn module takes only 28 lines of fairly simple code.

src/nn/visit_tensors.rs
```diff
@@ -0,0 +1,178 @@
...
+    }
+}
+
+#[non_exhaustive]
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum TensorVisitorOption {
+    DisableGradientUpdate,
+}
+
+pub trait TensorVisitor<const N: usize, const M: usize, E: Dtype, D: DeviceStorage> {
```
coreylowman, line 67: Is there anything preventing E/D generics being on the call method? Could simplify things a lot

src/nn/visit_tensors.rs
coreylowman: Ohhh I see, that's very clever!

src/nn/add_into.rs
```diff
@@ -23,12 +26,19 @@ use super::{BuildModule, BuildOnDevice, Module, ModuleMut, ResetParams, ToDevice
...
+        E: Dtype,
+        D: DeviceStorage,
+    > VisitTensorGroups<N, M, E, D> for AddInto<T>
+{
+    fn visit_groups<F: TensorVisitor<N, M, E, D>>(
+        mut self_refs: ModuleGroup<N, M, Self>,
+        func: &mut F,
+    ) -> Result<(), F::Err> {
+        self_refs.map(|s| &s.0, |s| &mut s.0, "0.").visit(func)
     }
```
coreylowman, line 42: Thoughts on including the visitor in the ModuleGroup object, and then call it inside .map?

If we can move the E/D generics to the call method, this shouldn't be an issue.

Something like:

```suggestion
    fn visit_groups<F: TensorVisitor<N, M>>(
        &mut self_refs: ModuleGroup<N, M, Self, F>,
    ) -> Result<(), F::Err> {
        self_refs.map(|s| &s.0, |s| &mut s.0, "0.")
    }
```

We could even go a step further and make ModuleGroup itself a TensorVisitor (wrapped around another TensorVisitor):

```suggestion
    fn visit_groups<V: TensorVisitor<N, M>>(
        &mut visitor: V,
    ) -> Result<(), F::Err> {
        visitor.map(|s| &s.0, |s| &mut s.0, "0.")
    }
```

src/nn/visit_tensors.rs
```diff
@@ -0,0 +1,178 @@
...
+{
+    fn visit_groups<F: TensorVisitor<N, M, E, D>>(
+        self_refs: ModuleGroup<N, M, Self>,
+        func: &mut F,
+    ) -> Result<(), F::Err> {
+        func.call(self_refs, &[])
+    }
+}
+
+pub trait VisitTensors<E: Dtype, D: DeviceStorage>: VisitTensorGroups<1, 0, E, D> + Debug {
```
coreylowman, line 97: Are `1, 0` and `0, 1` for VisitTensorsRef and VisitoTensorsMut the only two cases we would ever support?

I think we might be able to get rid of the VisitTensorGroups trait if we can separately impl some trait (not sure which one) for &T and &mut T

src/nn/visit_tensors.rs
nkoppel: Some TensorVisitors, including GradientUpdateVisitor, need to hold references to objects parametrized by a device and dtype, so moving these to call is not an option unfortunately.

src/nn/visit_tensors.rs
nkoppel: We will eventually support `1, 1` so we can add an multiply modules for EMA.

src/nn/add_into.rs
nkoppel: I've implemented your first suggestion and have renamed some things to make the new functionality of ModuleGroups make more sense. I haven't implemented your second suggestion because I don't really think I understand what you're getting at. What would the call implementation of ModuleGroup look like? How would we treat TensorVisitors not wrapped in a ModuleGroup?

coreylowman:
@nkoppel i figured out how to do it without the arrays, will open another PR soon

coreylowman:
Closing since the other PR was merged, thanks for the great work on this @nkoppel, the get_refs/get_muts is super clean/clever! 🚀

PR #469, opened 2023-02-21
Adds TensorCollection

### LINKED ISSUE #435: Parameter Count

coreylowman:
Resolves #435 
Related to #460 

- Removes trait GradientUpdate
- Removes trait ParamUpdater
- optimizers now implement VisitTensors and use RecursiveWalker in their Optimizer impl
- Moves trait ResetParams to use TensorCollection, and removes all nn impls
- Adds NumParams using TensorCollection
- Refactor SaveToNpz/LoadFromNpz to use TensorCollection
- Removes all missing gradients tests in nn
    - Now that all nn layers use TensorCollection, testing npz/update/reset_params is enough to cover module walking
    - Unused tensor tests are now added to optimizers to ensure they properly capture unused
- Adds traits VisitsTensors
- Adds trait ModuleVisitor
- Adds trait TensorCollection

TODOs
- [ ] add documentation
- [x] improve use statements
- [x] Move visitors somewhere else other than tensor? top level?


coreylowman:
@nkoppel thoughts? was able to impl optimizers with this approach as well

src/tensor/visitors/base.rs
```diff
@@ -0,0 +1,243 @@
...
+    {
+        self.path.push(opts.name.into());
+        self.f.visit(self.path.join("."), opts, get_muts(self.m))?;
+        self.path.pop();
+        Ok(())
+    }
+}
+
+impl<'a, M, E: Dtype, D: DeviceStorage, F: VisitTensorMutRef<E, D>> ModuleWalker<M, E, D>
+    for RecursiveWalker<'a, (&'a mut M, &'a M), F>
```
coreylowman, line 203: This impl is what enables EMA. Basically we'd just need to impl VisitTensorMutRef for some struct Ema and add an extension method

src/tensor/mod.rs
```diff
@@ -111,15 +111,13 @@
...
-
 #[cfg(feature = "cuda")]
 pub(crate) mod cuda;
-
 #[cfg(feature = "numpy")]
 pub(crate) mod numpy;
-
 pub(crate) mod storage_traits;
+mod tensor_impls;
+pub(crate) mod visitors; // TODO pub?
```
coreylowman, line 120: note to self

nkoppel: To me, the visitors module feels like it belongs in `nn`, because its main utility is in operating on nn modules.

coreylowman: That's what I did initially, however optim now also uses the visitors. optim depending on nn could makes sense the more I think about it...

src/tensor/visitors/reset_params.rs
```diff
@@ -0,0 +1,31 @@
...
+    }
+    fn try_reset_params(&mut self) -> Result<(), D::Err> {
+        Self::iter_tensors(&mut RecursiveWalker {
+            m: self,
+            f: &mut Resetter,
+            path: &mut Vec::new(),
+        })
+    }
+}
+impl<E: Dtype, D: DeviceStorage, M: TensorCollection<E, D>> ResetParams<E, D> for M {}
```
nkoppel, line 31: This and num_params should be moved to somewhere in `nn` to keep it discoverable in the documentation

src/tensor/visitors/base.rs
```diff
@@ -0,0 +1,243 @@
...
+    tensor::{DeviceStorage, Tensor},
+};
+
+use std::{string::String, vec::Vec};
+
+pub struct TensorOptions<S: Shape, E: Dtype, D: DeviceStorage> {
+    pub name: &'static str,
+    pub update: bool,
+    pub reset: fn(&mut Tensor<S, E, D>) -> Result<(), D::Err>,
+}
```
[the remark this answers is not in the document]
coreylowman: 👍 Yeah that's fair and good call out for consistency.

nkoppel:
One thing I don't like about the current implementation is the repeated code in `visitors/base.rs`. A lot of this repetition could be removed by generalizing the behavior of accessing fields in `&T`, `&mut T`, and `(&mut T, &T)` with something like the following:

```rust
pub trait VisitTensors<E: Dtype, D: DeviceStorage> {
    type Argument<'a, M>: VisitTensorArgument<M> where M: 'a;
    type Err;

    fn visit<S: Shape>(
        &mut self,
        full_path: String,
        opts: TensorOptions<S, E, D>,
        t: Self::Argument<'_, Tensor<S, E, D>>,
    ) -> Result<(), Self::Err>;
}

pub trait VisitTensorArgument<M> {
    type WithModule<'a, Mod>: VisitTensorArgument<Mod> where Self: 'a;

    fn get_field<Field, GetRef, GetMut>(
        &mut self,
        get_ref: GetRef,
        get_mut: GetMut,
    ) -> Self::WithModule<'_, Field>
     where
        GetRef: FnMut(&M) -> &Field,
        GetMut: FnMut(&mut M) -> &mut Field,
        Field: TensorCollection<E, D>,
}
```

This would remove the need to have three different `VisitTensor` traits, but would require that we add extra constraints to the `RecursiveWalker` struct to ensure that it has `F::Argument<M>` as it's `m` field. Thoughts?

coreylowman:
I totally agree, and nice sketch. Want to open a PR into this branch with that implemented? I think as long as we can keep that isolated to the implementation for RecursiveWalker (and not infect the user facing TensorCollection/TensorVisitor) it sounds like a great add

src/tensor/mod.rs
coreylowman: Moved

src/nn/mod.rs
```diff
@@ -104,6 +104,10 @@
 //! mlp.load_state_dict(state_dict)
 //! ```
 
+mod num_params;
+mod reset_params;
+pub mod visitors;
```
coreylowman, line 109: Come up with a better name than "visitors". The real purpose for users will be getting access to TensorCollection to impl for modules. Perhaps "collections"?

src/nn/visitors.rs
```diff
@@ -0,0 +1,255 @@
...
+        get_muts: GetMut,
+        name: &str,
+        opts: TensorOptions<S, E, D>,
+    ) -> Result<(), Self::Err>
+    where
+        GetRef: FnMut(&T) -> &Tensor<S, E, D>,
+        GetMut: FnMut(&mut T) -> &mut Tensor<S, E, D>;
+}
+
+pub(crate) struct RecursiveWalker<'a, M, F> {
```
coreylowman, line 121: This should probably be `pub`. Anyone who wants to implement their own optimizer outside of dfdx will need access to this

coreylowman:
@nkoppel I'm pretty happy with where this is, any other organization/naming stuff you think we should do before merge?

src/nn/tensor_collection/visitor.rs
```diff
@@ -0,0 +1,198 @@
...
+
+    /// Return the view of the tensor
+    fn view<'a, Mod, Field, GetRef, GetMut>(
+        module: &'a mut Self::View<'_, Mod>,
+        get_ref: &mut GetRef,
+        get_mut: &mut GetMut,
+    ) -> Self::View<'a, Field>
+    where
+        GetRef: FnMut(&Mod) -> &Field,
+        GetMut: FnMut(&mut Mod) -> &mut Field;
```
nkoppel, line 47: I like using 'view' to express the meaning of this trait, but the naming/documentation of the view function isn't very informative.
```suggestion
    /// Given a view to a module, returns a view to one of that module's fields
    fn view_field<'a, Mod, Field, GetRef, GetMut>(
        module: &'a mut Self::View<'_, Mod>,
        get_ref: &mut GetRef,
        get_mut: &mut GetMut,
    ) -> Self::View<'a, Field>
    where
        GetRef: FnMut(&Mod) -> &Field,
        GetMut: FnMut(&mut Mod) -> &mut Field;
```

src/nn/tensor_collection/collection.rs
```diff
@@ -0,0 +1,96 @@
...
+        )
+    }
+}
+
+/// Options to change behavior of [TensorVisitor]
+pub struct TensorOptions<S: Shape, E: Dtype, D: DeviceStorage> {
+    /// Whether the tensor should be updated with gradients
+    pub do_gradient_update: bool,
+    pub reset: fn(&mut Tensor<S, E, D>) -> Result<(), D::Err>,
+}
```
nkoppel, line 63: I think we should add `#[non_exhaustive]` to this struct so that we can easily add new options in the future if needed. We could also define default values for all fields and use the builder pattern to construct this, but I'm not sure this is better.

nkoppel:
@usbalbin merging this will break parts of #437. You'll needto remove the GradientUpdate implementation and replace it with a TensorCollection implementation.

src/nn/tensor_collection/collection.rs
coreylowman: Oh just kidding, anyone implementing Visitors will need to access those fields. I'll do #[non_exhaustive]!

PR #471, opened 2023-02-21
Rename to `add_dim`.

infalmo:


coreylowman:
@infalmo looks good. What are your thoughts on changing it to `trait PushDim` and `fn push_dim`? Otherwise will merge

infalmo:
`AddDim` is good. `PushDim` sounds like we're pushing the dimension, which doesn't make sense tbh.

PR #472, opened 2023-02-21
Add TensorContainer trait to allow more argument types for TensorVisitors in #469

nkoppel:
Adds TensorContainer to allow implementers of VisitTensors to specify their own argument types. This is done with the `Container` associated type, which represents some collection of immutable and mutable references to the same type. References to a particular type are represented with `&'static ()` and `&'static mut ()`, but these could be substituted for other types including unconstructable enums, like these:
```rust
enum Ref {}
enum RefMut {}
```
Related to #469 and #460.


src/nn/visitors.rs
```diff
@@ -124,131 +123,51 @@ pub(crate) struct RecursiveWalker<'a, M, F> {
...
     {
         self.path.push(name.into());
         let mut walker = RecursiveWalker {
-            m: (get_muts(self.m.0), get_refs(self.m.1)),
+            m: F::Container::get_field(&mut self.m, &mut get_refs, &mut get_muts),
             f: self.f,
             path: self.path,
         };
         Field::iter_tensors(&mut walker)?;
+        std::mem::drop(walker);
```
coreylowman, line 149: Is this necessary?

src/nn/npz.rs
```diff
@@ -112,10 +112,12 @@ pub trait LoadFromNpz<E: Dtype + NumpyDtype, D: CopySlice<E>>: TensorCollection<
 }
 impl<E: Dtype + NumpyDtype, D: CopySlice<E>, T: TensorCollection<E, D>> LoadFromNpz<E, D> for T {}
 
-impl<W: Write + Seek, E: Dtype + NumpyDtype, D: CopySlice<E>> VisitTensorRef<E, D>
+impl<W: Write + Seek, E: Dtype + NumpyDtype, D: CopySlice<E>> VisitTensors<E, D>
     for zip::ZipWriter<W>
 {
+    type Container = &'static ();
```
coreylowman, line 118: Yeah lets go with your suggestion in the PR description about using the unconstructable enums instead of these, purely for readability. I know i'd forget in a couple months what this means haha. 

```rust
type Container = TensorRef;
```
makes so much sense!

src/nn/impl_tensor_container.rs
```diff
@@ -0,0 +1,95 @@
...
+        GetMut: FnMut(&mut Mod) -> &mut Field,
+    {
+        module
+            .iter_mut()
+            .map(|x| T::get_field(x, get_ref, get_mut))
+            .collect()
+    }
+}
+
+impl<T: TensorContainer> TensorContainer for Option<T> {
```
coreylowman, line 81: I could see this being used for optional fields (e.g. bias if we go the Option route), nice!

coreylowman (approved):
This is awesome!!! Going to merge for now and change to the unconstructable enums (& do some reorganization in the branch)

src/nn/visitors.rs
nkoppel: Yes, if you remove it, rust will consider `self.path` to be mutably borrowed on the next line.

PR #482, opened 2023-02-24
Add try_forward and try_forward_mut to Module and ModuleMut.

### LINKED ISSUE #323: Fallible Module & ModuleMut
- [ ] Add `Module::try_forward` and `ModuleMut::try_forward_mut`
- [ ] Start impling for things
- [ ] Make forward default to try_forward.unwrap

nkoppel:
Closes #323. I've tried my best to convert every operation in implementations of `try_forward` and `try_forward_mut` to their `try_` counterpart, but it's possible that I have missed some.

### DIFF
--- a/src/nn/batchnorm2d.rs
+++ b/src/nn/batchnorm2d.rs
@@ -82,9 +82,9 @@ impl<const C: usize, E: Dtype, D: Device<E>> BatchNorm2D<C, E, D> {
         let mean = self.running_mean.clone();
 
         // normalize & affine
-        let x = x.try_sub(mean.broadcast_like(&shape))?;
-        let x = x.try_div(std.broadcast_like(&shape))?;
-        let x = x.try_mul(self.scale.clone().broadcast_like(&shape))?;
+        let x = x.try_sub(mean.try_broadcast_like(&shape))?;
+        let x = x.try_div(std.try_broadcast_like(&shape))?;
+        let x = x.try_mul(self.scale.clone().try_broadcast_like(&shape))?;
         x.try_add(self.bias.clone().try_broadcast_like(&shape)?)
     }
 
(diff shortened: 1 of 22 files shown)


src/nn/batchnorm2d.rs
```diff
@@ -71,67 +71,81 @@ pub struct BatchNorm2D<const C: usize, E: Dtype, D: DeviceStorage> {
...
         let mean = self.running_mean.clone();
 
         // normalize & affine
-        let x = sub(x, mean.broadcast_like(&shape));
-        let x = div(x, std.broadcast_like(&shape));
-        let x = mul(x, self.scale.clone().broadcast_like(&shape));
-        add(x, self.bias.clone().broadcast_like(&shape))
+        let x = x.try_sub(mean.broadcast_like(&shape))?;
+        let x = x.try_div(std.broadcast_like(&shape))?;
+        let x = x.try_mul(self.scale.clone().broadcast_like(&shape))?;
```
coreylowman, line 87:
```suggestion
        let x = x.try_sub(mean.try_broadcast_like(&shape))?;
        let x = x.try_div(std.try_broadcast_like(&shape))?;
        let x = x.try_mul(self.scale.clone().try_broadcast_like(&shape))?;
```

src/nn/batchnorm2d.rs
```diff
@@ -300,7 +326,7 @@ mod tests {
     }
 
     #[test]
-    fn test_batchform2d_3d_repeated_forward_mut() {
+    fn test_batchnorm2d_3d_repeated_forward_mut() {
```
coreylowman, line 329: 🤣

coreylowman (approved):
This looks good to me, nice work! Re the try_broadcast, I wonder if we should remove the fallible version of that (and with permute) since we know they won't allocate new memory. This would make them inconsistent with the other tensor ops though...

nkoppel:
In my opinion, we should keep the fallible versions of broadcast and permute, because users shouldn't be exposed to the fact that broadcast and permute can't error when writing fallible code.

PR #484, opened 2023-02-24
Optimizing cpu kernels for reductions

coreylowman:
This moves cpu kernels closer to what cuda kernels do for reductions:
1. Permute the shape/strides so that all reduced elements are rightmost dims
2. Iterate over out buffers in order (so no nd-indexing)
3. Advance the input iter (nd indexed) by the number of elements reduced

Additionally this adds some shortcuts for full reductions (to 0d)

### DIFF
--- a/src/tensor_ops/utilities/reduction_utils.rs
+++ b/src/tensor_ops/utilities/reduction_utils.rs
@@ -1,7 +1,47 @@
-#[cfg(feature = "cuda")]
-use crate::prelude::{Axes, Shape};
+use crate::shapes::{Axes, Shape};
+use crate::tensor::cpu::NdIndex;
 #[cfg(feature = "cuda")]
 use std::vec::Vec;
+
+/// Permutes strides so that all reduces axes are rightmost.
+/// For example reducing (2, 3, 4) to (4,) would permute the shape to (4, 2, 3)
+/// via the strides.
+///
+/// This makes all the reduced elements sequential when indexed using [NdIndex].
+#[inline(always)]
+pub(crate) fn index_for_reductions<S: Shape, Ax: Axes>(
+    shape: S,
+    strides: S::Concrete,
+) -> NdIndex<S> {
+    let dims = shape.concrete();
+    let mut new_shape: S::Concrete = Default::default();
+    let mut new_strides: S::Concrete = Default::default();
+    let num_non_reduced_dims = S::NUM_DIMS - Ax::as_array().into_iter().count();
+
+    let mut i_reduced = 0;
+    let mut i_non_reduced = 0;
+    for i_src in 0..S::NUM_DIMS {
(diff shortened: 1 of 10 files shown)


src/tensor_ops/utilities/reduction_utils.rs
```diff
@@ -1,8 +1,48 @@
...
 #[cfg(feature = "cuda")]
 use std::vec::Vec;
 
+/// Permutes strides so that all reduces axes are rightmost.
+/// For example reducing (2, 3, 4) to (4,) would permute the shape to (4, 2, 3)
+/// via the strides.
+///
+/// This makes all the reduced elements sequential when indexed using [NdIndex].
+#[inline(always)]
+pub(crate) fn index_for_reductions<S: Shape, Ax: Axes>(
```
coreylowman, line 12: @nkoppel i'm pretty sure this does exactly what the cuda version below does. Thoughts on if we could combine them? I'd like to keep the cpu one not using vecs, which I think is the biggest difference.

src/tensor_ops/utilities/reduction_utils.rs
```diff
@@ -1,8 +1,48 @@
...
+    let num_non_reduced_dims = S::NUM_DIMS - Ax::as_array().into_iter().count();
+
+    let mut i_reduced = 0;
+    let mut i_non_reduced = 0;
+    for i_src in 0..S::NUM_DIMS {
+        if Ax::as_array().into_iter().any(|x| x == i_src as isize) {
+            // this axis is reduced
+            new_shape[num_non_reduced_dims + i_reduced] = dims[i_src];
+            new_strides[num_non_reduced_dims + i_reduced] = strides[i_src];
+            i_reduced += 1;
```
coreylowman, line 28: An optimization for this is probably to sort the strides in this section of `new_shape`/`new_strides` by decreasing size, to maximize locality when indexing.

src/tensor_ops/utilities/reduction_utils.rs
```diff
@@ -1,8 +1,48 @@
...
+            i_non_reduced += 1;
+        }
+    }
+    NdIndex {
+        indices: Default::default(),
+        shape: new_shape,
+        strides: new_strides,
+        next: Some(0),
+        contiguous: (new_shape == dims && new_strides == shape.strides())
+            .then(|| shape.num_elements()),
```
coreylowman, line 42: Add a unit test for this logic, for things like .softmax() on the last axis, this should be true

src/tensor_ops/utilities/reduction_utils.rs
nkoppel: I think we can factor out the shared logic from index_for_reductions and permute_for_reductions, but I'd prefer to keep the filtering of broadcasted dimensions in permute_for_reductions.

This implementation looks correct to me, and is likely faster than my implementation, so go ahead and use yours.

PR #497, opened 2023-02-26
Adding attention_reshape (inference only) kernels.

Narsil:
Tried to implement `cat` but there were some weird issues, and it was 
causing a ton of kernels being called, with each some device to host sync (huge bottleneck).

This PR adds a pretty specific reshape for `(qkv, past_key, past_value)` useful for
inference run of transformer models. For instance `gpt2` (https://github.com/Narsil/fast_gpt2)


Narsil:
@coreylowman

coreylowman (approved):
Nice contribution, thanks! 🚀

PR #503, opened 2023-02-28
Allowing `nn::Embedding` to be dynamic in shape.

Narsil:
This is important for transformers at inference where the sequence size
is usually unknown at runtime (Depends on the tokenized string).

I took the liberty of putting the `BATCH` also as dynamical as it's the
same in inference, there's usually no way to know in advance (for
webserver kind of dynamic batching).


Narsil:
@coreylowman

PR #504, opened 2023-02-28
Adding `UnbiasedLinear` (linear without bias).

Narsil:
- Propose to add a different layer. It's uncommon enough that I don't
  think polluting `Linear` with bias being `Option<Tensor>` should be
ok.
- But it does come in handy for the LanguageModel head within
  `transformers`. I don't have a lot of other useful examples tbh.

In theory we could use a regular Linear and just init the bias to zero,
but that's a waste of operations afterwards.

### DIFF
--- a/src/nn/unbiased_linear.rs
+++ b/src/nn/unbiased_linear.rs
@@ -1,0 +1,286 @@
+use crate::{gradients::Tape, shapes::*, tensor::*, tensor_ops::*};
+
+use super::{tensor_collection::*, BuildModule, BuildOnDevice, Module, NonMutableModule, ToDevice};
+
+use num_traits::Float;
+use rand_distr::{uniform::SampleUniform, Uniform};
+
+pub mod builder {
+    #[derive(Debug, Copy, Clone, Eq, PartialEq)]
+    pub struct UnbiasedLinear<const I: usize, const O: usize>;
+}
+
+impl<const I: usize, const O: usize, E: Dtype, D: Device<E>> BuildOnDevice<D, E>
+    for builder::UnbiasedLinear<I, O>
+where
+    UnbiasedLinear<I, O, E, D>: BuildModule<D, E>,
+{
+    type Built = UnbiasedLinear<I, O, E, D>;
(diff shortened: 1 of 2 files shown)


Narsil:
@coreylowman

src/nn/unbiased_linear.rs
```diff
@@ -0,0 +1,286 @@
...
+    type Error = D::Err;
+
+    /// 1d forward using [matmul()] and [add()].
+    fn try_forward(&self, x: T) -> Result<Self::Output, D::Err> {
+        x.try_matmul(self.weight.retaped::<T::Tape>().try_permute()?)
+    }
+}
+
+#[derive(Clone, Debug)]
+struct Bias1D<'a, const M: usize, E: Dtype, D: DeviceStorage> {
```
coreylowman, line 110: You can remove this since this doesn't have bias 😀

src/nn/unbiased_linear.rs
```diff
@@ -0,0 +1,286 @@
...
+        }
+    }
+}
+
+impl<const I: usize, const O: usize, E: Dtype, D: Device<E>, T> Module<T>
+    for UnbiasedLinear<I, O, E, D>
+where
+    T: SplitTape + TryMatMul<Tensor<Rank2<I, O>, E, D, T::Tape>> + HasErr<Err = D::Err>,
+    T::Tape: Tape<D>,
+    for<'a> Bias1D<'a, O, E, D>: Module<T::Output, Output = T::Output, Error = D::Err>,
```
coreylowman, line 98: You can remove this bound

coreylowman (changes requested):
Looks good, after removal of the Bias1D struct we can merge!

src/nn/unbiased_linear.rs
Narsil: Done !

src/nn/unbiased_linear.rs
Narsil: Done !

coreylowman (approved):
Nice!!

PR #505, opened 2023-02-28
Making K dimension of matmul dynamic.

Narsil:
This one is a bit more controversial I think.

Within `transformers` models, in generative mode (`gpt2`, `chatGPT`,
`whisper` etc..), there is a need for `past_keys_values` which is
a cache for the attention layer, reducing the size of matmuls.

The one matmul where something happens is `matmul(qk, v)`.

Q is of shape (NUM_HEADS, SEQ, HEAD_DIM)
K is of shape (NUM_HEADS, HEAD_DIM, PAST + SEQ)
So

QK is of shape (NUM_HEADS, SEQ, PAST + SEQ)

V is of shape (NUM_HEADS, PAST + SEQ, HEAD_DIM)

Where SEQ is the sequence_length of the input string.
      PAST is the sequence of the cached attention.

Therefore we here have a matrix multiplication where the inner dimension
(denoted K in dfdx traits) is actually dynamic and not compile time static.

This PR only modifies 1 kernel, but probably all kernels should be
modified if that is deemed worthy. But doing so might actually defeat
the purpose of getting compile time static shapes. (Does it though?
there should still be errors at compile time.)

My personal take/understanding, is that it would be OK to implement
everything dynamic for `try_kernel` functions, and keep the `kernel`
only for the ones that cannot fail.
However, there are other issues like striding that could cause fail on
ConstShape kernels too (rendering `kernel` functions potentially
panicking).

### DIFF
--- a/src/tensor_ops/matmul/cpu_kernel.rs
+++ b/src/tensor_ops/matmul/cpu_kernel.rs
@@ -1,5 +1,5 @@
 use crate::shapes::*;
-use crate::tensor::cpu::{Cpu, StridedArray, View, ViewMut};
+use crate::tensor::cpu::{Cpu, CpuError, StridedArray, View, ViewMut};
 
 #[cfg(not(feature = "cblas"))]
 use matrixmultiply::{dgemm, sgemm};
@@ -235,13 +235,19 @@ impl<F: Dtype> super::MatMatBatch3Kernel<F> for Cpu
 where
     Self: MatMulImpl<F>,
 {
-    fn forward<const B: usize, M: Dim, const K: usize, N: Dim>(
+    fn forward<const B: usize, M: Dim, K: Dim, N: Dim>(
         &self,
-        lhs: &Self::Storage<(Const<B>, M, Const<K>), F>,
-        rhs: &Self::Storage<(Const<B>, Const<K>, N), F>,
+        lhs: &Self::Storage<(Const<B>, M, K), F>,
+        rhs: &Self::Storage<(Const<B>, K, N), F>,
     ) -> Result<Self::Storage<(Const<B>, M, N), F>, Self::Err> {
         let m: M = lhs.shape().1;
         let n: N = rhs.shape().2;
+
+        let k: K = lhs.shape().2;
+        let k2: K = rhs.shape().1;
+        if k != k2 {
+            return Err(CpuError::WrongNumElements);
+        }
         let mut out = StridedArray::new((Const, m, n))?;
         let a = lhs.view();
         let b = rhs.view();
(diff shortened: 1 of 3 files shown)


Narsil:
@coreylowman

src/tensor_ops/matmul/cpu_kernel.rs
```diff
@@ -235,13 +235,19 @@ impl<F: Dtype> super::MatMatBatch3Kernel<F> for Cpu
...
+        rhs: &Self::Storage<(Const<B>, K, N), F>,
     ) -> Result<Self::Storage<(Const<B>, M, N), F>, Self::Err> {
         let m: M = lhs.shape().1;
         let n: N = rhs.shape().2;
+
+        let k: K = lhs.shape().2;
+        let k2: K = rhs.shape().1;
+        if k != k2 {
+            return Err(CpuError::WrongNumElements);
+        }
```
coreylowman, line 250: you can just do `assert_eq!` here instead of returning err. All the shape checks we have so far panic immediately instead of returning an err (the err is for memory allocation mainly)

coreylowman:
> This PR only modifies 1 kernel, but probably all kernels should be
modified if that is deemed worthy. But doing so might actually defeat
the purpose of getting compile time static shapes. (Does it though?
there should still be errors at compile time.)

I think as we move more and more of the kernels to accept both const/dyn, I'm okay with this. Users of dynamic shapes will have to accept that there will runtime panics, but when using const shapes, you'll still get an error.

So we can move all of the matmul kernels to dynamic.

We should add some doctests that are annotated as `no_compile` to ensure this

Narsil:
> Users of dynamic shapes will have to accept that there will runtime panics,

With some kind of `Dyn<'B'>` there would also be compile errors. /inception.

Narsil:
> We should add some doctests that are annotated as no_compile to ensure this

Not sure what kind of coverage we would want here. 
There's probably Vec Vs Mat vs MatBr vs MatBack vs Mat4d and possibly 2 tests for each (1 for when the input have a mismatch in the K dimension, and 1 for when the receiver has a dimension mismatch, does  it make sense ?)

coreylowman:
> Not sure what kind of coverage we would want here.
There's probably Vec Vs Mat vs MatBr vs MatBack vs Mat4d and possibly 2 tests for each (1 for when the input have a mismatch in the K dimension, and 1 for when the receiver has a dimension mismatch, does it make sense ?)

Yeah ideally all the methods, except vecvec i guess.

For run time panics (ie. usizes) we can add `#[test]` that are also annotated with `#[should_panic = "<reason>"]`

For compile time panics we'll need to add a doctest for each one.

I'd say we can add the checks & doctest on the kernel impls in mod.rs. For example:

```diff
impl<
        const K: usize,
        N: Dim,
        E: Dtype,
        D: VecMatKernel<E>,
        T: Tape<E, D> + Merge<R>,
        R: Tape<E, D>,
    > TryMatMul<Tensor<(Const<K>, N), E, D, R>> for Tensor<(Const<K>,), E, D, T>
{
    type Output = Tensor<(N,), E, D, T>;
    /// ```no_compile
    /// ...
    /// ```
    fn try_matmul(self, rhs: Tensor<(Const<K>, N), E, D, R>) -> Result<Self::Output, Self::Err> {
        assert_eq!(self.shape.0, rhs.shape.0);
        try_binary_op(self, rhs, D::forward, D::backward)
    }
}
```

coreylowman:
Awesome, thanks for the contribution!

PR #511, opened 2023-03-03
Adding axpy tensor op & ModelEMA module walker

### LINKED ISSUE #425: Add EMA (exponential moving average) support for NN

coreylowman:
Resolves #425


src/tensor_ops/ema/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+    tensor::{DeviceStorage, Tensor},
+};
+
+mod cpu_kernel;
+#[cfg(feature = "cuda")]
+mod cuda_kernel;
+
+/// Exponential moving average: `dst * decay + src * (1 - decay)`. See [ema_assign]
+/// for in place version.
+pub fn ema<S: Shape, E: Unit, D>(
```
coreylowman, line 12: Should change this to be a more general: `c = alpha * a + beta * b`.

Then this could actually be used for ema (alpha = decay, beta = 1 - decay), and gradient accum (alpha = 1, beta = 1).

src/nn/ema.rs
```diff
@@ -0,0 +1,117 @@
...
+}
+
+/// Performs model exponential moving average on two modules.
+pub trait ModelEMA<E: Dtype, D: EmaKernel<E>>: TensorCollection<E, D> {
+    /// Does `self = self * decay + other * (1 - decay), using
+    /// [crate::tensor_ops::ema] on parameters.
+    ///
+    /// **Only updates trainable parameters**. For example, batch normalization
+    /// running parameters are not updated.
+    fn ema_assign(&mut self, other: &Self, decay: E) {
```
coreylowman, line 36: TODO come up with a better name than this? Is `ema` fine? Do we need something to imply that it is in place?

PR #513, opened 2023-03-03
feat: adds BatchNorm1D

### LINKED ISSUE #476: Add BatchNorm1D
Implementation of BatchNorm2D exists, but of BatchNorm1D doesn't.

kstavro:
Closes #476.

@coreylowman Unfortunately, I couldn't deduplicate the code. As mentioned in the issue, capturing the infer and train logic inside a BatchNorm trait complained about all trys (tried to implement `TryAdd` and `TrySub` at many places, but the compiler kept complaining  regardless of my efforts), whereas leaving the infer and train logic inside functions has the problem that the return is no longer a `Result`, but a tuple (because we have to return `running_mean` and `running_var` as well). The `?` operator complained a lot about it.

This should be fine for the moment.  Both 2d and 3d variants are there. I also touched the respective docu a bit. Performance hit on 06-MNIST is ~1.5-1.8x when adding batchnorm layers. Would like to add some stuff to the MNIST-example in another issue, maybe option for batchnorm can be one of them.


kstavro:
Ok, that didn't go well. Weird, all batchnorm tests passed locally. I wonder if it has to do with the fact that my fork was some commits behind... Will check it again later.

coreylowman:
Yeah just happened to update the tape interface after a while. Should be stable now though!

coreylowman:
@kstavro this should be possible with functions. But notably the running_mean/running_var can be `&mut` and updated inside the function, instead of returned.

Here's the function headers that would work:

```rust
fn train_fwd<const C: usize, S: Shape, E: Dtype, D: Device<E>, T: Tape<E, D>, Ax: Axes>(
    x: Tensor<S, E, D, T>,
    var: &mut Tensor<Rank1<C>, E, D>,
    mean: &mut Tensor<Rank1<C>, E, D>,
    scale: &Tensor<Rank1<C>, E, D>,
    bias: &Tensor<Rank1<C>, E, D>,
    epsilon: E,
    momentum: E,
) -> Result<Tensor<S, E, D, T>, D::Err>
where
    S: HasAxes<Ax> + ReduceShapeTo<Rank1<C>, Ax>,
{
    ...
}

fn infer_fwd<const C: usize, S: Shape, E: Dtype, D: Device<E>, Ax: Axes>(
    x: Tensor<S, E, D>,
    var: &Tensor<Rank1<C>, E, D>,
    mean: &Tensor<Rank1<C>, E, D>,
    scale: &Tensor<Rank1<C>, E, D>,
    bias: &Tensor<Rank1<C>, E, D>,
    epsilon: E,
) -> Result<Tensor<S, E, D>, D::Err>
where
    Rank1<C>: BroadcastShapeTo<S, Ax>,
{
    ...
}
```

and then you could invoke like:

```rust
        train_fwd(
            x,
            &mut self.running_var,
            &mut self.running_mean,
            &self.scale,
            &self.bias,
            self.epsilon,
            self.momentum,
        )
```

kstavro:
Yeah, apart from returning the `running_mean` and the `running_var` as well, the function headers were exactly the same.

I was under the impression that if you don't return the running constants, you wouldn't be able to update the attributes of the struct (wanted to destructure the return of train to update the attributes). But probably you are right.

Then let me try it tomorrow. Where should I pack the functions? Maybe a separate file called `batchnorm.rs`?

coreylowman:
Since we pass them as `&mut`, they can be updated by the function. Luckily rust is smart enough to allow us to mutable borrow multiple fields of a struct at the same time!

As far as where, let's just stick them in either of the batchnorm1d or batchnorm2d file, and then the other one can just do a `use super::batchnorm2d::{infer_fwd, train_fwd}`

kstavro:
Any tips on how to assign the new `mean` and `var` values inside the `train_fwd`? I had actually tried this before attemting to return them at the end of `train_fwd` and the compiler complains a lot, whatever variation of the assignemt `mean = &mut mean.clone()....` I have used. 

If there was a way to Deref a Tensor, I could use `*mean = *mean.clone()...`, so I went and implemented Deref for Tensor but this broke the whole repo because of auto-recursion limits being violated (no idea what this means in Rust), so that was not an option in the end. I knew it wouldn't work, because it would have already been implemented, but still tried it out out of curiosity.

Next thing I tried was:

```rust
let temp_mean = &mut mean
        .clone()
        .try_mul(E::ONE - momentum)?
        .try_add(mean_chan.retaped::<NoneTape>().try_mul(momentum)?)?;
mean.data = temp_mean.data; // directly assigning mean = temp_mean was worse
```

but data is wrapped inside Arc and this caused: `cannot move out of `temp_mean.data` which is behind a mutable reference
move occurs because `temp_mean.data` has type `Arc<<D as tensor::storage_traits::DeviceStorage>::Vec<E>>`, which does not implement the `Copy` trait`.

Just sharing my progress so far in case you can provide any fast tips of how to get around this. Will revisit it tomorrow.

kstavro:
@coreylowman not sure what I did there, I just merged the changes from main into my fork. merged the batchnorm conflicts and repushed in my fork. I guess I need to "recontribute" the code?

kstavro:
Just tried it locally: I think `std::mem::replace(mean, ...);` also works for the update of the running variables, but not sure which one is preferable in terms of memory footprint/performance. And some sort of `#[warn(unused_must_use)]` might be needed in that case, because the replacement is not assigned to a new variable and the compiler gives a warning (I tried `std::mem::take`, but apparently `Tensor` needs to implement `Default` for that.)

coreylowman:
Ahh I see the issue. A recent change might have made this a bit easier actually - the latest batchnorm2d uses `try_axpy` to do the updates to running_mean/running_var in place. If you move to that you shouldn't have to do the derefs

kstavro:
Should be good now. Quick question: does the default `DeviceStorage` now implement `AxpyKernel`? Couldn't find that in code directly.

coreylowman:
Device<E> does, DeviceStorage does not. Device<E> is just DeviceStorage + a toooon of kernels for that dtype

coreylowman (approved):
Looks perfect, nice addition 😀

Issue #490, opened 2023-02-25
Should binary ops invoke two separate kernels for backward?


coreylowman:
Related to #487 

Could be an optimization to launch two separate concurrent kernels for binary op backwards

coreylowman:
This could be parallelized with streams as well similar to #492

PR #499, opened 2023-02-27
WIP Considering broadcasted elements in binop backwards

coreylowman:
Related to #491

PR #520, opened 2023-03-04
Use chunk_sum in cuda kernels of backward binary operations.

### LINKED ISSUE #487: Reduce usages of atomicAdd in binary op backward for cuda kernels
Currently this exists to cover cases when the input was broadcasted. However not every tensor is broadcasted, so this may be able to only call atomicAdd for tensors that are.

nkoppel:
Resolves #487 and #490. This should substantially improve performance of batched training on cuda, as the backward time of the conv2d benchmark measures about 1.5 times faster on my system. 

We call two kernels because grad_out needs to be permuted in two different ways so that we can sum the gradients using chunk_sum regardless of which dimensions of lhs and rhs are broadcasted.


kstavro:
@nkoppel how cuda specific are these optimizations? Are they easy to implement for/translate to CPU as well? In the batchnorm1d PR, `BatchNorm1D` layers have a ~1.5x performance hit and I would be into trying to translate in a new PR your changes into a CPU version of chunk_sum, etc (if you think this is possible/makes sense).

nkoppel:
While it doesn't really make sense to implement chunk_sum on the cpu, the techniques used in this pr may speed up the backward of binary operations on the cpu. "chunk_sum" is a *parallel* reduction algorithm, meaning that it depends on having access to hundreds or thousands of cores at one time to be fast, and therefore can't be implemented on the cpu. 

However, using [permute_for_binary_backward](https://github.com/coreylowman/dfdx/pull/520/files#diff-e1a368ef4528a0ecc570492cedc67c3789616f59067e7f19f80002894fb96d81) may improve performance by changing the order in which the elements of tensors are read so that all values that need to be summed for either grad_lhs or grad_rhs end up next to each other. This reordering will likely improve performance despite the number of cache-unfriendly reads remaining constant, because each pass will only need to do two cache-unfriendly reads. This may allow the cpu to use its cache more effectively.

Also note that @coreylowman may already be working on this in #499.

kstavro:
Got it, so you would only look into better cache management instead of also trying to use all cores of the CPU (even if they are only 8 compared to the thousands of cuda). Thanks for the heads up!

Will stay put on the issue for the moment and also keep an eye on #499. Will check the code from both PRs just to see how you guys do it (not that I am any good with cuda, but you have to start from somewhere).

coreylowman:
Nice improvement! thanks for the contribution 🚀

PR #524, opened 2023-03-07
Adding `model.alloc_grads()`, removing `Default` for `Gradients`

### LINKED ISSUE #523: Out of memory issues with gradient accumulation
This is because the temporary gradients are kept around in the gradients object.

Is it possible to have gradients object clear temporary gradients?

We could do something like:

```rust
// would set some flags on which gradients are not temporary
let grads: Gradients<_, _> = model.alloc_grads();
...
let grads = loss.backward(); // backward would clear the temporary gradients after running
```

coreylowman:
Resolves #523

### DIFF
--- a/src/gradients.rs
+++ b/src/gradients.rs
@@ -1,7 +1,7 @@
 //! Implementations of [OwnedTape], [NoneTape], and generic Nd array containers via [Gradients].
 #![allow(clippy::type_complexity)]
 
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::{boxed::Box, vec::Vec};
 
 use crate::shapes::{Shape, Unit};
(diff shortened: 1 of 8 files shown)


src/gradients.rs
```diff
@@ -52,7 +54,15 @@ impl<E: Unit, D: DeviceStorage> Gradients<E, D> {
 
     /// Drops all gradients except for the ids specified in the parameter.
     pub fn retain(&mut self, ids: &[UniqueId]) {
-        self.gradient_by_id.retain(|k, _| ids.contains(k));
+        self.non_temporary.extend(ids);
```
coreylowman, line 57: non_temporary needs to be a set so this doesn't keep accumulating the same ids. A set may also be faster to query!

src/gradients.rs
```diff
@@ -52,7 +54,15 @@ impl<E: Unit, D: DeviceStorage> Gradients<E, D> {
...
+        self.drop_temporaries();
+    }
+
+    /// Keeps all gradients marked previously by [Gradients::retain], and drops all
+    /// others.
+    pub fn drop_temporaries(&mut self) {
+        self.gradient_by_id
+            .retain(|k, _| self.non_temporary.contains(k));
```
coreylowman, line 65: Needs to handle when retain hasn't been called (i.e. all gradients are temporary)

PR #529, opened 2023-03-08
Fixing bool tests with safetensors (serde compatibility)

coreylowman:
@Narsil apparently serde messes with the bool tests:

```rust
error[E0283]: type annotations needed
   --> src\tensor_ops\boolean\mod.rs:285:51
    |
285 |         assert_eq!(r3.array(), dev.zeros_like(&a).array());
    |         ------------------------------------------^^^^^--- type must be known at this point
    |
    = note: multiple `impl`s satisfying `bool: PartialEq<_>` found in the following crates: `core`, `serde_json`:
            - impl PartialEq for bool;
            - impl PartialEq<serde_json::value::Value> for bool;
```

Weird error, I'm not exactly sure why rust thinks serde_json::value::Value is a possible for zeros_like.


Narsil:
Okay thanks !

PR #530, opened 2023-03-08
Adds `Tensor::concat`

### LINKED ISSUE #43: Add `concatenate` function
Needed for #34 . In multi head attention, you concatenate the output of all the single attention heads.

Note that this may require nightly access similar to #1 because we can't do expressions with const generics yet.

### LINKED ISSUE #463: Add `tensor::cat` method

coreylowman:
Resolves #43
Resolves #463


coreylowman:
Thanks for the help @infalmo 🚀

PR #531, opened 2023-03-08
Hotfixing the safetensors impl.

Narsil:
- Checking the actual shapes before loading the data.
- Requires changing the error types since this is specific type of
  error.


Narsil:
@coreylowman I just checked and I missed some failures. This PR should fix that.

coreylowman (approved):
👍

PR #538, opened 2023-03-08
Handle path for TensorVisitors using a TensorViewer

nkoppel:
Removes the full_path parameter from TensorVisitors and adds the `ViewTensorName` TensorViewer to allow SaveToNpz and LoadFromNpz to still receive each tensor's name. This slightly reduces overhead when using operations which don't require tensor names, and slightly increases overhead when storing and loading tensors.


coreylowman:
Black magic 🤯 will take a closer look later tonight. TensorCollection/TensorVisitor is turning into my favorite part of dfdx lol. It's too damn useful

coreylowman (approved):
Super useful, iterating tensors is getting simpler and simpler 🚀

Issue #475, opened 2023-02-22
Add ToDtype trait for converting between dtypes


coreylowman:
This should be similar to ToDevice where you can call `tensor.to_dtype::<f32>()`.

Similarly, we can implement for things in nn `module.to_dtype::<f32>()`.

This will require implementing some cuda kernels.

Ideally we should also look some ways to simplify implementation at the nn level, since at the moment we still have to specify ToDtype/ToDevice explicitly for each struct

PR #548, opened 2023-03-10
Allow Modules to be constructed with the TensorCollection trait

nkoppel:
Expands the definition of TensorCollection to allow it to build Modules to allow implementing BuildModule, ToDevice, and ToDtype with TensorCollection. To facilitate this, TensorVisitor has been split in TensorFunction and TensorVisitor, and usize is no longer a Dtype.

Tasks:
- [ ] Implement BuildModule
- [ ] Implement ToDevice
- [ ] Implement ToDtype (#475)

### DIFF
--- a/src/nn/add_into.rs
+++ b/src/nn/add_into.rs
@@ -1,4 +1,4 @@
-use crate::{shapes::Dtype, tensor::*};
+use crate::{prelude::Device, shapes::Dtype, tensor::*};
 
 use super::traits::*;
 
@@ -33,9 +33,17 @@ impl<T: BuildModule<D, E>, D: DeviceStorage, E: Dtype> BuildModule<D, E> for Add
     }
 }
 
-impl<E: Dtype, D: DeviceStorage, T: TensorCollection<E, D>> TensorCollection<E, D> for AddInto<T> {
-    fn iter_tensors<V: ModuleVisitor<Self, E, D>>(visitor: &mut V) -> Result<(), V::Err> {
-        visitor.visit_module("0", |s| &s.0, |s| &mut s.0)
+impl<E: Dtype, D: Device<E>, T: TensorCollection<E, D>> TensorCollection<E, D> for AddInto<T> {
+    type Output<E2: Dtype, D2: Device<E2>> = AddInto<T::Output<E2, D2>>;
+
+    fn iter_tensors<V: ModuleVisitor<Self, E, D>>(
+        visitor: &mut V,
+    ) -> ModuleVisitorOutput<V::Func, Self, E, D> {
+        Ok(Some(AddInto(crate::try_option!(visitor.visit_module(
+            "0",
+            |s| &s.0,
+            |s| &mut s.0
+        )?))))
     }
 }
 
--- a/src/nn/batchnorm1d.rs
+++ b/src/nn/batchnorm1d.rs
@@ -5,6 +5,8 @@ use super::{
     traits::*,
 };
 
+use num_traits::FromPrimitive;
+
 pub mod builder {
     #[derive(Debug, Copy, Clone, Eq, PartialEq)]
     pub struct BatchNorm1D<const C: usize>;
@@ -181,31 +183,44 @@ impl<const C: usize, E: Dtype, D: Device<E>> BuildModule<D, E> for BatchNorm1D<C
 }
 
 impl<const C: usize, E: Dtype, D: Device<E>> TensorCollection<E, D> for BatchNorm1D<C, E, D> {
-    fn iter_tensors<V: ModuleVisitor<Self, E, D>>(visitor: &mut V) -> Result<(), V::Err> {
-        visitor.visit_tensor(
+    type Output<E2: Dtype, D2: Device<E2>> = BatchNorm1D<C, E2, D2>;
+
+    fn iter_tensors<V: ModuleVisitor<Self, E, D>>(
+        visitor: &mut V,
+    ) -> ModuleVisitorOutput<V::Func, Self, E, D> {
+        let scale = visitor.visit_tensor(
             "scale",
             |s| &s.scale,
             |s| &mut s.scale,
             TensorOptions::reset_to_ones(),
         )?;
-        visitor.visit_tensor(
+        let bias = visitor.visit_tensor(
             "bias",
             |s| &s.bias,
             |s| &mut s.bias,
             TensorOptions::reset_to_zeros(),
         )?;
-        visitor.visit_tensor(
+        let running_mean = visitor.visit_tensor(
             "running_mean",
             |s| &s.running_mean,
             |s| &mut s.running_mean,
             TensorOptions::detached(|t| t.try_fill_with_zeros()),
         )?;
-        visitor.visit_tensor(
+        let running_var = visitor.visit_tensor(
             "running_var",
             |s| &s.running_var,
             |s| &mut s.running_var,
             TensorOptions::detached(|t| t.try_fill_with_ones()),
-        )
+        )?;
+
+        Ok(Some(BatchNorm1D {
+            scale: crate::try_option!(scale),
+            bias: crate::try_option!(bias),
+            running_mean: crate::try_option!(running_mean),
+            running_var: crate::try_option!(running_var),
+            epsilon: <V::Func as TensorFunction<E, D>>::OutE::from_f32(1e-5).unwrap(),
+            momentum: <V::Func as TensorFunction<E, D>>::OutE::from_f32(0.1).unwrap(),
+        }))
     }
 }
 
--- a/src/nn/conv.rs
+++ b/src/nn/conv.rs
@@ -64,8 +64,12 @@ where
     E: Dtype + Float + SampleUniform,
     D: Device<E>,
 {
-    fn iter_tensors<V: ModuleVisitor<Self, E, D>>(visitor: &mut V) -> Result<(), V::Err> {
-        visitor.visit_tensor(
+    type Output<E2: Dtype, D2: Device<E2>> = Conv2D<I, O, K, S, P, E2, D2>;
+
+    fn iter_tensors<V: ModuleVisitor<Self, E, D>>(
+        visitor: &mut V,
+    ) -> ModuleVisitorOutput<V::Func, Self, E, D> {
+        let weight = visitor.visit_tensor(
             "weight",
             |s| &s.weight,
             |s| &mut s.weight,
@@ -73,7 +77,11 @@ where
                 let b = E::ONE / E::from_usize(I * K * K).unwrap().sqrt();
                 t.try_fill_with_distr(rand_distr::Uniform::new(-b, b))
             }),
-        )
+        )?;
+
+        Ok(Some(Conv2D {
+            weight: crate::try_option!(weight),
+        }))
     }
 }
 
--- a/src/nn/ema.rs
+++ b/src/nn/ema.rs
@@ -2,12 +2,12 @@ use super::tensor_collection::{
     RecursiveWalker, TensorCollection, TensorOptions, TensorVisitor, ViewTensorMut, ViewTensorRef,
 };
 
-use crate::{shapes::*, tensor::*, tensor_ops::axpy::AxpyKernel};
+use crate::{prelude::Device, shapes::*, tensor::*};
 
 struct ModelEMAOp<E> {
     decay: E,
 }
-impl<E: Dtype, D: AxpyKernel<E>> TensorVisitor<E, D> for ModelEMAOp<E> {
+impl<E: Dtype, D: Device<E>> TensorVisitor<E, D> for ModelEMAOp<E> {
     type Viewer = (ViewTensorMut, ViewTensorRef);
     type Err = D::Err;
 
@@ -24,7 +24,7 @@ impl<E: Dtype, D: AxpyKernel<E>> TensorVisitor<E, D> for ModelEMAOp<E> {
 }
 
 /// Performs model exponential moving average on two modules.
-pub trait ModelEMA<E: Dtype, D: AxpyKernel<E>>: TensorCollection<E, D> {
+pub trait ModelEMA<E: Dtype, D: Device<E>>: TensorCollection<E, D> {
     /// Does `self = self * decay + other * (1 - decay), using
     /// [crate::tensor_ops::axpy()] on parameters.
     ///
@@ -39,10 +39,11 @@ pub trait ModelEMA<E: Dtype, D: AxpyKernel<E>>: TensorCollection<E, D> {
         Self::iter_tensors(&mut RecursiveWalker {
             m: (self, other),
             f: &mut op,
-        })
+        })?;
+        Ok(())
     }
 }
-impl<E: Dtype, D: AxpyKernel<E>, M: TensorCollection<E, D>> ModelEMA<E, D> for M {}
+impl<E: Dtype, D: Device<E>, M: TensorCollection<E, D>> ModelEMA<E, D> for M {}
 
 #[cfg(test)]
 mod tests {
--- a/src/nn/module.rs
+++ b/src/nn/module.rs
@@ -1,9 +1,9 @@
-use crate::shapes::Dtype;
 #[cfg(feature = "cuda")]
 pub use crate::tensor::OnCuda;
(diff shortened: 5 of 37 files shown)


src/nn/tensor_collection/collection.rs
```diff
@@ -1,28 +1,62 @@
...
 /// that operates on tensors, like resetting, EMA, counting number of params,
 /// gradient updates, etc.
-pub trait TensorCollection<E: Dtype, D: DeviceStorage>: Sized {
-    fn iter_tensors<V: ModuleVisitor<Self, E, D>>(visitor: &mut V) -> Result<(), V::Err>;
+pub trait TensorCollection<E: Dtype, D: Device<E>>: Sized {
+    type Output<E2: Dtype, D2: Device<E2>>;
+
+    fn iter_tensors<V: ModuleVisitor<Self, E, D>>(
+        visitor: &mut V,
+    ) -> ModuleVisitorOutput<V::Func, Self, E, D>;
```
coreylowman, line 45: I guess to support ToDtype & ToDevice we'd need different generics for dtype & device here?
Something like?
```suggestion
    fn iter_tensors<E2, D2, V: ModuleVisitor<Self, E2, D2>>(
        visitor: &mut V,
    ) -> ModuleVisitorOutput<V::Func, Self, E2, D2>;
```

coreylowman:
Yeah I like this a lot, it'd be great to unify all of these together. Will be watching how this evolves as you merge todevice/build module in.

Also curious if we can simplify the option handling. Mixing options/results is such a pain haha

src/nn/tensor_collection/collection.rs
nkoppel: Yes, I was using associated types of TensorFuncitons to specify the output dtype and device, but I have moved over to using type parameters.

src/nn/batchnorm1d.rs
```diff
@@ -167,61 +167,45 @@ impl<B: Dim, const C: usize, E: Dtype, D: Device<E>>
...
-    fn to_device(&self, device: &D2) -> Self::Output {
-        BatchNorm1D {
-            scale: self.scale.to_device(device),
-            bias: self.bias.to_device(device),
-            running_mean: self.running_mean.to_device(device),
-            running_var: self.running_var.to_device(device),
-            epsilon: self.epsilon,
-            momentum: self.momentum,
-        }
+        Ok(crate::try_some!(BatchNorm1D {
```
coreylowman, line 201: is `crate::try_some` just doing this?
```suggestion
        Ok(Some(BatchNorm1D {
```

nkoppel: The main purpose of try_some is to wrap that code in a closure that returns an Option, so that the ? operator can be used to propagate Nones. We need to use the ? operator twice on the results so that we short circuit any errors immediately, while not skipping over any tensors when we aren't outputting a Module.

examples/07-custom-module.rs
```diff
@@ -1,49 +1,64 @@
...
+        // not be handled until we construct the output Mlp, because a value of None is not an error.
+        let l1 = visitor.visit_module("l1", |s| &s.l1, |s| &mut s.l1)?;
+        let l2 = visitor.visit_module("l2", |s| &s.l2, |s| &mut s.l2)?;
+
+        // Specify how to construct Mlp given optional values for its fields.
+        Ok(dfdx::try_some!(Mlp {
+            l1: l1?,
+            l2: l2?,
+            relu: Default::default(),
+        }))
```
coreylowman, line 50: I think the mixing of option/result may be a bit too subtle for average user. I.e. it might be too easy to mess this method up.

We should explore some alternative UX options for this. It'd be nice to have something like a builder pattern here, where users would have to always call .build() at the end:

```rust
visitor
    .module("l1", ...)
    .module("l2", ...)
    .build(|(l1, l2)| Mlp { l1, l2 relu: Default::default })
```

If we could get to something *like* that, I think we could avoid the users having to think about options/results at all, and instead handle them in the background.

I know this would be a large departure from what we have now, but it'd be a big improvement IMO

nkoppel: Yeah, this is a much nicer syntax, and while your suggestion doesn't seem to be possible, it should be possible to have the following syntax:
```rust
visitor.visit_fields(
    (
        ModuleField("l1", |s| &s.l1, |s| &mut s.l1),
        ModuleField("l2", |s| &s.l2, |s| &mut s.l2),
        TensorField("example", |s| &s.example, |s| &mut s.example, TensorOptions::...),
    ),
    |(l1, l2, example)| {
        Mlp {
            l1,
            l2,
            example,
            relu: Default::default()
        }
    }
)
```
using something like the following trait:
```rust
pub trait ModuleFields<M, E: Dtype, D: Device<E>> {
    type Options<E2, D2>;
    type Output<E2, D2>;

    fn visit_fields<E2: Dtype, D2: Device, V: ModuleVisitor<M, E, D, E2, D2>>(
        &mut self,
        module: M
    ) -> Result<Self::Options<E2, D2>, Self::Err>;

    fn handle_options<E2: Dtype, D2: Device>(Self::Options<E2, D2>) -> Option<Self::Output<E2, D2>>;
}
```

coreylowman: Yeah I tried for a bit yesterday to do the builder pattern, but it was very difficult 😅 

I *really* like that, nice idea! We might even be able to get rid of the visitor all together, and just have the TensorCollection with a static method that returns the fields & build closure:

```rust
   let fields = (
        ModuleField("l1", |s| &s.l1, |s| &mut s.l1),
        ModuleField("l2", |s| &s.l2, |s| &mut s.l2),
        TensorField("example", |s| &s.example, |s| &mut s.example, TensorOptions::...),
    );
    let build = |(l1, l2, example)| {
        Mlp {
            l1,
            l2,
            example,
            relu: Default::default()
        }
    };
    (fields, build)
```

nkoppel: That's a really great idea, but it seems impractical, because this would either require specifying the entire type of both `fields` and `build` or trait objects, because you cannot return `impl Trait` from trait methods.

I think that the builder pattern is technically possible, but it would require something like TupleLists to function, which is less than ideal.

src/nn/ema.rs
```diff
@@ -1,25 +1,23 @@
...
+use super::tensor_collection::*;
 
-use crate::{shapes::*, tensor::*, tensor_ops::axpy::AxpyKernel};
+use crate::{prelude::Device, shapes::*, tensor::*};
 
 struct ModelEMAOp<E> {
     decay: E,
 }
-impl<E: Dtype, D: AxpyKernel<E>> TensorVisitor<E, D> for ModelEMAOp<E> {
+impl<E: Dtype, D: Device<E>> TensorVisitor<E, D, f32, Cpu> for ModelEMAOp<E> {
```
coreylowman, line 8:
```suggestion
impl<E: Dtype, D: Device<E>> TensorVisitor<E, D, E, D> for ModelEMAOp<E> {
```

src/nn/build_module.rs
```diff
@@ -0,0 +1,43 @@
...
+    type Err = D::Err;
+
+    fn visit<S: Shape>(
+        &mut self,
+        opts: TensorOptions<S, E, D>,
+        _t: (),
+    ) -> Result<Option<Tensor<S, E, D>>, Self::Err> {
+        let mut tensor: Tensor<S, E, D> = self.0.try_zeros_like(&opts.shape)?;
+        (opts.reset)(&mut tensor)?;
+        Ok(Some(tensor))
```
coreylowman, line 16: 🔥

src/nn/add_into.rs
```diff
@@ -23,26 +23,20 @@ use super::*;
...
 
-impl<T: ToDevice<D>, D> ToDevice<D> for AddInto<T> {
-    type Output = AddInto<T::Output>;
-    fn to_device(&self, device: &D) -> Self::Output {
-        AddInto(self.0.to_device(device))
+    fn iter_tensors<E2: Dtype, D2: Device<E2>, V: ModuleVisitor<Self, E, D, E2, D2>>(
+        visitor: &mut V,
+    ) -> Result<Option<Self::Output<E2, D2>>, V::Err> {
+        visitor.visit_fields(
+            ModuleField::new("0", |s: &Self| &s.0, |s| &mut s.0),
```
coreylowman, line 37: I think we can get around having to specify `&Self` on these by adding a default method in the TensorCollection trait:

```rust
trait TensorCollection {
   fn iter_tensors(...);
    fn module(...) -> ModuleField { ... }
    fn tensor(...) -> TensorField { ... }
}
```

and then we could call like:
```suggestion
            Self::module("0", |s| &s.0, |s| &mut s.0),
```

src/nn/conv.rs
```diff
@@ -64,47 +64,22 @@ where
...
     D: Device<E>,
 {
-    fn iter_tensors<V: ModuleVisitor<Self, E, D>>(visitor: &mut V) -> Result<(), V::Err> {
-        visitor.visit_tensor(
+    type Output<E2: Dtype, D2: Device<E2>> = Conv2D<I, O, K, S, P, E2, D2>;
+
+    fn iter_tensors<E2: Dtype, D2: Device<E2>, V: ModuleVisitor<Self, E, D, E2, D2>>(
+        visitor: &mut V,
+    ) -> Result<Option<Self::Output<E2, D2>>, V::Err> {
+        let weight = visitor.visit_tensor(
```
coreylowman, line 72: Still old version?

src/nn/module.rs
```diff
@@ -79,26 +75,17 @@ impl<D: DeviceStorage> DeviceBuildExt for D {}
...
+    type Output<E2: Dtype, D2: Device<E2>> = T;
 
-impl<T: ZeroSizedModule + Clone, D> ToDevice<D> for T {
-    type Output = T;
-    fn to_device(&self, _device: &D) -> Self {
-        self.clone()
+    fn iter_tensors<E2: Dtype, D2: Device<E2>, V: ModuleVisitor<Self, E, D, E2, D2>>(
+        _visitor: &mut V,
+    ) -> Result<Option<Self::Output<E2, D2>>, V::Err> {
+        Ok(Some(Default::default()))
```
coreylowman, line 88: Should this call `visitor.visit_fields((), Default::default)`?

src/nn/num_params.rs
```diff
@@ -1,21 +1,21 @@
 use super::tensor_collection::*;
 
-use crate::{shapes::*, tensor::*};
+use crate::{prelude::Device, shapes::*, tensor::*};
 
 struct Counter(usize);
-impl<E: Dtype, D: DeviceStorage> TensorVisitor<E, D> for Counter {
+impl<E: Dtype, D: Device<E>> TensorVisitor<E, D, f32, Cpu> for Counter {
```
coreylowman, line 6:
```suggestion
impl<E: Dtype, D: Device<E>> TensorVisitor<E, D, E, D> for Counter {
```

src/nn/tensor_collection/collection.rs
```diff
@@ -1,71 +1,156 @@
...
 
+use super::{visitor::TensorVisitor, ModuleFields};
+
+#[macro_export]
+macro_rules! try_some {
+    {$e:expr} => {{
+        let f = move || {Some($e)};
+        f()
+    }};
+}
```
coreylowman, line 17: Can this be removed now?

src/nn/tensor_collection/collection.rs
```diff
@@ -1,71 +1,156 @@
...
+///
+/// let dev = Cpu::default();
+/// let model = Mlp::<f32, Cpu>::build(&dev);
+/// assert_eq!(132, model.num_trainable_params());
+///
+/// ```
+pub trait TensorCollection<E: Dtype, D: Device<E>>: Sized {
+    /// Type alias that specifies the how a module's type changes when using a different dtype and/or
+    /// device.
+    type Output<E2: Dtype, D2: Device<E2>>;
```
coreylowman, line 62: Thoughts on calling this `To` instead of `Output`?

src/nn/tensor_collection/collection.rs
```diff
@@ -1,71 +1,156 @@
...
-pub struct TensorOptions<S: Shape, E: Dtype, D: DeviceStorage> {
+pub struct TensorOptions<S: Shape, E: Dtype, D: Device<E>> {
     /// Whether the tensor should be updated with gradients
     pub do_gradient_update: bool,
 
     /// How to reset the tensor in the future.
     pub reset: fn(&'_ mut Tensor<S, E, D>) -> Result<(), D::Err>,
+
+    /// The [Shape] that BuildModule uses to construct the tensor
+    pub shape: S,
```
coreylowman, line 150: How about Option<S>? Defaults of usize dims will be 0 which may lead to some subtle bugs

src/nn/tensor_collection/visitor.rs
```diff
@@ -16,16 +15,18 @@ pub struct RecursiveWalker<'a, M, F> {
 }
 
 /// Something that can visit [Tensor]s. Used in conjunction with [RecursiveWalker].
-pub trait TensorVisitor<E: Dtype, D: DeviceStorage> {
+pub trait TensorVisitor<E: Dtype, D: Device<E>, E2: Dtype, D2: Device<E2>> {
```
coreylowman, line 18: I was trying yesterday to figure out how we could move this E2 & D2 out of the trait and capture it in some `TensorVisitor::Output` associated type, but failed to.

Any ideas how we could do that? It would greatly simplify some of the trait definitions/generics if we only needed 1 set. We'd basically only need to handle a second dtype/device in the ToDevice/ToDtype impls of TensorVisitor if we could figure that out

coreylowman (commented):
Looking great!

src/nn/conv.rs
nkoppel: Resolved with 5cb574d

src/nn/tensor_collection/collection.rs
nkoppel: I don't know. I'd like to keep it so that we don't have to specify the shape of tensors with ConstShapes, but I don't see a way to do this without trait specialization.

src/nn/tensor_collection/visitor.rs
nkoppel: I've actually decided to make E2 and D2 associated types of both ModuleVisitor and TensorVisitor to make referring to them when implementing iter_tensors easier.

coreylowman: Great work!

src/nn/tensor_collection/collection.rs
```diff
@@ -74,6 +187,7 @@ impl<S: Shape, E: Dtype, D: DeviceStorage> TensorOptions<S, E, D> {
         TensorOptions {
             do_gradient_update: true,
             reset: |t| t.try_fill_with_zeros(),
+            shape: S::const_default(),
```
coreylowman, line 190: Another option here is to only allow creating TensorOptions for `S: ConstShape`, which does impl Default. I'd prefer this than adding Default bounds to Shape & Dim.

src/shapes/shape.rs
```diff
@@ -65,19 +65,23 @@ pub trait Dtype:
...
 
 /// Represents something that has a [Dtype].
 pub trait HasDtype {
     type Dtype: Dtype;
 }
 
 /// Represents a single dimension of a multi dimensional [Shape]
-pub trait Dim: 'static + Copy + Clone + std::fmt::Debug + Send + Sync + Eq + PartialEq {
+pub trait Dim:
+    'static + Copy + Clone + std::fmt::Debug + Send + Sync + Eq + PartialEq + Default
```
coreylowman, line 82: I had left Default off of both Dim & Shape purposely - mainly for runtime dimensions. While nothing in dfdx depends on Default impl for shapes anymore, I could see someone using this in a generic context, which would be misleading.

I think you can probably revert these changes to shape, and instead just use ConstShape for TensorOptions like I mention above. This does limit all the tensors to be compile time known for nn, but that's what they are for the foreseeable future anyway.

If we change nn to support runtime dimensions we can address this there. I suspect it'll require a lot of api changes anyway

coreylowman (changes requested):
Super close, I think just update the handling of the shapes and then this looks good to merge! 🚀 🚀 🚀

coreylowman:
Great work @nkoppel, ready to merge?

PR #557, opened 2023-03-11
Moving transformers to stable, and accepting dyn dimensions for transformer input

### LINKED ISSUE #554: Compilation hanging with dynamic dimensions
I've been switching my language model project over to using dyn dims for batch size and seq len, and for an unknown reason compilation hangs and leaks memory. I also get an error when running rust-analyzer due to an nvcc error, but that's a seperate concern.

I've created a reproduction here: https://github.com/jafioti/dfdx_issue

Just clone it and try to build.

coreylowman:
Resolves #554


coreylowman:
FYA @Narsil @jafioti

PR #561, opened 2023-03-12
feat: add realize shape

### LINKED ISSUE #532: Add a way to convert between const/dyn shapes
For example:

```rust
let a: Tensor<Rank2<2, 3>, f32, _> = dev.zeros();
let a: Tensor<(usize, usize), f32, _> = a.realize_shape().unwrap();
let a: Tensor<Rank2<2, 3>, f32, _> = a.realize_shape().unwrap();
```

1. The number of dimensions must match between the two shapes, and thus the following conversions should make sense. This can be specified by doing `Dst: Shape<Concrete = S::Concrete>`.
2. You can use `Dst::from_concrete` to convert the shapes.
3. The method should return an Option, in case the shape's concretes don't match.

Internally, this can just move everything on the tensor, something like:

```rust
impl<...> Tensor<...> {
    pub fn realize_shape<Dst: Shape>(self) -> Option<Tensor<Dst, ...>> {
        let shape = Dst::from_concrete(self.shape.concrete())?
       Some(Tensor { ... shape, ... })
    }
}
```

Alexandcoats:
# Description

This PR adds the ability to coerce shapes and tensors freely between their const and dynamic representions.

## Related Issues

- Closes #532


src/tensor_ops/realize_to.rs
```diff
@@ -0,0 +1,213 @@
...
+        Ok(self.shape.realized().map(|dst_shape| Tensor {
+            id: self.id,
+            data: self.data,
+            strides: dst_shape.strides(),
+            shape: dst_shape,
+            device: self.device,
+            tape: self.tape,
+        }))
+    }
+}
```
nkoppel, line 40: It should be fine to do something like the following unless you want to have an error type to describe the specific error:

```suggestion
impl<S: Shape, E: Dtype, D: DeviceStorage, T: Tape<E, D>> RealizeTo for Tensor<S, E, D, T> {
    fn try_realize<Dst: Shape>(self) -> Result<Result<Self::WithShape<Dst>, Self>, Self::Err>
    where
        Self::Shape: RealizeShapeTo<Dst>,
    {
        if let Some(dst_shape) = self.shape.realized() {
            Ok(Ok(Tensor {
                id: self.id,
                data: self.data,
                strides: dst_shape.strides(),
                shape: dst_shape,
                device: self.device,
                tape: self.tape,
            }))
        } else {
            Ok(Err(self))
        }
    }
}
```

src/tensor_ops/realize_to.rs
```diff
@@ -0,0 +1,213 @@
...
+
+impl<S: Shape, E: Dtype, D: DeviceStorage, T: Tape<E, D>> RealizeTo for Tensor<S, E, D, T> {
+    fn try_realize<Dst: Shape>(self) -> Result<Option<Self::WithShape<Dst>>, Self::Err>
+    where
+        Self::Shape: RealizeShapeTo<Dst>,
+    {
+        Ok(self.shape.realized().map(|dst_shape| Tensor {
+            id: self.id,
+            data: self.data,
+            strides: dst_shape.strides(),
```
nkoppel, line 34: Also, this needs to have the same value as self.strides, because `dst_shape.strides()` is not necessarily the same as the input strides when the input is broadcasted/permuted.

src/tensor_ops/realize_to.rs
Alexandcoats: It's strange to have the ability to specify an error type and simply refer back to one that's so limiting and generic. My preference would be to have an error type for this trait, but I wasn't sure what @coreylowman's thoughts were on it.

coreylowman: TBH I don't even think we need try_realize. The main point of try_* right now is to catch memory allocation errors or other device level errors (like launching cuda kernel errors). Since realize doesn't re-allocate, or really do anything at the device level, it won't ever fail like this.

src/tensor_ops/realize_to.rs
coreylowman: Yep agree, this can just be
```suggestion
            strides: self.strides,
```

src/tensor_ops/realize_to.rs
Alexandcoats: In that case, should `realize` return a `Result<_, Self>` or something similar to allow recovering from a failed call?

coreylowman: Yeah I like that idea!

src/tensor_ops/realize_to.rs
```diff
@@ -0,0 +1,214 @@
...
+    /// or returns the original tensor if the new shape's dimensions are incompatable.
+    /// ```rust
+    /// # use dfdx::prelude::*;
+    /// # let dev: Cpu = Default::default();
+    /// let a: Tensor<Rank2<2, 3>, f32, _> = dev.zeros();
+    /// let a = a.realize::<(usize, usize)>().unwrap();
+    /// let mut a = a.realize::<Rank2<2, 3>>().unwrap();
+    /// match a.realize::<(usize, Const<4>)>() {
+    ///     Ok(new) => handle_realized(new),
+    ///     Err(old) => handle_failure(old),
```
coreylowman, line 15: For the sake of simple example we can just do a print statement:
```suggestion
    ///     Ok(new) => println!("Shape was properly realized, returned new tensor"),
    ///     Err(old) => println!("Shape could not be realized, returned the original tensor"),
```

Alexandcoats: Thanks!

coreylowman:
Thanks for the contribution! 🚀

PR #566, opened 2023-03-13
Removes .trace_into(), .trace() now requires Gradients object

### LINKED ISSUE #565: Users should allocate gradients instead of allocating them with `trace()`
_Originally posted by @coreylowman in https://github.com/coreylowman/dfdx/issues/523#issuecomment-1466817341_
            
Currently `x.trace()` creates a `Gradients::without_leafs()` object. This can cause huge memory consumption because the leaf tensors are kept around until the gradients object is destroyed, or `model.zero_grads` is called with the gradients.

Additionally, it also means you can write training code without explicitly creating a gradients object, meaning you'll be allocating & de-allocating model gradients for each batch, which is **very** expensive.

Instead, dfdx should require users to pre-allocate gradients, and trace should be changed to accept a gradients object. While this would make simple tensor operations a little less ergonomic, it will make for a better user experience in the long run because people will have better performance.

Additionally, it makes debugging large memory usage easier because if someone is using `Gradients::without_leafs`, it's likely that.

This would look like:

```rust
let mut grads = model.alloc_grads();
...
x.trace(grads);
```

coreylowman:
Resolves #565


src/nn/dropout.rs
```diff
@@ -91,7 +93,7 @@ impl<const N: usize, S: Shape, E: Dtype, D: Device<E>> ModuleMut<Tensor<S, E, D,
 /// # use dfdx::prelude::*;
 /// # let dev: Cpu = Default::default();
 /// let dropout: Dropout = Default::default();
-/// dropout.forward(dev.zeros::<Rank1<5>>().trace());
+/// dropout.forward(dev.zeros::<Rank1<5>>().trace_all());
```
coreylowman, line 96: Fix

PR #582, opened 2023-03-18
Add ToDtype tensor operation

### LINKED ISSUE #475: Add ToDtype trait for converting between dtypes

nkoppel:
Adds to_dtype method for tensors, and will add a TensorVisitor implementation to convert models between dtypes. "to_dtype" is implemented for most conversions between Units, using AsPrimitive on Cpu and macros on Cuda. Due to the large amount of generated code, compiling test_to_dtype adds around 3 seconds to compile times for both cuda and cpu on my machine.

Currently, to_dtype can not be used with OwnedTapes because Gradients can only hold data with a single dtype.

Closes #475.

Tasks:

- [x] Add TensorVisitor in nn
- [x] Documentation

### DIFF
--- a/src/tensor_ops/to_dtype/mod.rs
+++ b/src/tensor_ops/to_dtype/mod.rs
@@ -1,0 +1,64 @@
+mod cpu_kernel;
+#[cfg(feature = "cuda")]
+mod cuda_kernel;
+
+use crate::prelude::{DeviceStorage, Shape, Tensor, Unit};
+
+pub trait ToDtypeKernel<E1: Unit, E2: Unit>: DeviceStorage {
+    fn forward<S: Shape>(inp: Tensor<S, E1, Self>) -> Result<Tensor<S, E2, Self>, Self::Err>;
+}
+
+impl<S: Shape, E: Unit, D: DeviceStorage> Tensor<S, E, D> {
+    pub fn try_to_dtype<E2: Unit>(self) -> Result<Tensor<S, E2, D>, D::Err>
+    where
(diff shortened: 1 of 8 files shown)


src/tensor_ops/to_dtype/mod.rs
```diff
@@ -0,0 +1,87 @@
...
+        self.try_to_dtype().unwrap()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::{prelude::Rank1, tensor::*, tests::*};
+
+    #[test]
+    fn test_to_dtype() {
```
coreylowman, line 52: So just this is increasing compile times by that much? I guess because it needs a generic version for every combo...

Is it possible to reduce the times by splitting this up into separate test functions? I'm not sure how rustc compiles test but maybe it compiles separate tests in parallel?

nkoppel: That does seem to have improved compile times on my machine, but I'm not sure that it will have a benefit for github actions, as I don't think they are given access to multiple threads.

coreylowman: Lucky us 😀

PR #598, opened 2023-03-22
Better cuda conv2d kernel

### LINKED ISSUE #578: Optimize conv2d folding kernels
Currently in convolution networks, the unfold_output & unfold_input kernels take a signficant portion of time. This issue is about optimizing them to reduce this time.

My guess is there are some things we can do with multi-dimensional block/grid dims, as well as reducing the number of branches in the kernel.

opfromthestart:
This optimizes the Conv2D kernel.
Some of these changes should be reverted as they were just how I benchmarked my changes.
I tested this on a GTX 1650 Ti and it improved from taking 171.5 seconds to 86.6 seconds.
Fixes #578

### DIFF
--- a/src/tensor_ops/conv2d/conv2d.cu
+++ b/src/tensor_ops/conv2d/conv2d.cu
@@ -41,19 +41,23 @@ __device__ void unfold_input_into_patches(
 
     const size_t y_plus_p = oh * op.stride + k1;
     if (y_plus_p < op.padding) {
+        patches[i] = 0;
         return;
     }
     const size_t y = y_plus_p - op.padding;
     if (y >= op.h_in) {
+        patches[i] = 0;
         return;
     }
 
     const size_t x_plus_p = ow * op.stride + k2;
     if (x_plus_p < op.padding) {
+        patches[i] = 0;
         return;
     }
     const size_t x = x_plus_p - op.padding;
     if (x >= op.w_in) {
+        patches[i] = 0;
         return;
     }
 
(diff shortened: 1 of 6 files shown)


src/tensor_ops/conv2d/conv2d.cu
```diff
@@ -78,43 +79,50 @@ __device__ void unfold_output_into_patches(
...
+            ow -= k2;
+            if (ow % op.stride != 0) {
+                return;
+            }
+            ow /= op.stride;
+            if (ow >= op.w_out) {
+                return;
+            }
+
+            size_t patch_i = b * (op.chan_out * op.kernel * op.kernel * op.h_out * op.w_out) +
```
opfromthestart, line 117: Some of this calculation could be moved outside the loop, but if its optimized right I don't think it would make a difference.

coreylowman:
@opfromthestart this seems like it fails cuda tests - you can run conv2d tests with:
```rust
cargo +nightly test tensor_ops::conv2d -F test-cuda
```

opfromthestart:
@coreylowman I had an error conflating input and output dimensions. It should be fixed now, as those 6 tests pass. Could cuda be added to the automatic tests?

coreylowman:
@opfromthestart unfortunately no because github actions doesn't have access to a GPU. There may be a way to set up a specific machine to run tests on gpu, but I haven't looked into it in depth

src/tensor_ops/conv2d/cuda_kernel.rs
```diff
@@ -102,13 +102,14 @@ where
         _: &Tensor<O, E, Self>,
         grad_out: &Self::Vec<E>,
     ) -> Result<(), Self::Err> {
-        let patches_numel = op.batch * op.chan_out * op.kernel * op.kernel * op.h_in * op.w_in;
+        let image_numel = op.batch * op.chan_out * op.h_in * op.w_in;
+        let patches_numel = op.kernel * op.kernel * image_numel;
         let mut patches = self.dev.alloc_zeros::<E>(patches_numel)?;
```
coreylowman, line 107: Any thoughts on how to get rid of this memset operation here? I tried changing this to `self.dev.alloc::<E>`, which just allocates but doesn't do the memset, but the unfolding kernel doesn't necessarily set every position if its launched on the image pixels as you do below.

If we can get rid of the memsets on the patches its a pretty big improvement too

opfromthestart: I changed all the alloc_zeros to alloc except for `grad_f_b1023`, and it didnt seem to make a difference (82.5s alloc vs 86.6s alloc_zeros, which is within variance). I'm guessing memset is run async (or its just fast) and so it doesn't actually affect the runtime much.

Issue #591, opened 2023-03-20
`reshape` does not work on tensors of type `usize`


Cifram:
When I attempt to run this line:
```
let upsample_gather: Tensor<Rank1<{ SIZE * 2 }>, usize, Cuda> = upsample_gather.reshape();
```
I get this error:
```
error[E0599]: the method `reshape` exists for struct `Tensor<(Const<SIZE>, Const<2>), usize, Cuda>`, but its trait bounds were not satisfied
  --> src\upsample.rs:28:83
   |
28 |         let upsample_gather: Tensor<Rank1<{ SIZE * 2 }>, usize, Cuda> = upsample_gather.reshape();
   |                                                                                         ^^^^^^^ method cannot be called on `Tensor<(Const<SIZE>, Const<2>), usize, Cuda>` due to unsatisfied trait bounds
   |
  ::: C:\Users\heruf\.cargo\registry\src\index.crates.io-6f17d22bba15001f\dfdx-0.11.0\src\tensor\tensor_impls.rs:32:1
   |
32 | pub struct Tensor<S: Shape, E: Unit, D: DeviceStorage, T = NoneTape> {
   | -------------------------------------------------------------------- doesn't satisfy `_: ReshapeTo`
   |
  ::: C:\Users\heruf\.cargo\registry\src\index.crates.io-6f17d22bba15001f\dfdx-0.11.0\src\tensor\cuda\device.rs:14:1
   |
14 | pub struct Cuda {
   | --------------- doesn't satisfy `_: ReshapeKernel<usize>`
   |
   = note: the following trait bounds were not satisfied:
           `usize: Dtype`
           which is required by `dfdx::tensor::Tensor<(Const<SIZE>, Const<2>), usize, dfdx::tensor::Cuda>: dfdx::tensor_ops::ReshapeTo`
           `dfdx::tensor::Cuda: tensor_ops::reshape_to::ReshapeKernel<usize>`
           which is required by `dfdx::tensor::Tensor<(Const<SIZE>, Const<2>), usize, dfdx::tensor::Cuda>: dfdx::tensor_ops::ReshapeTo`
```
Before that line, `upsample_gather` has type `Tensor<(Const<SIZE>, Const<2>), usize, Cuda>`.

[...]

PR #602, opened 2023-03-22
Add Slice tensor operation.

### LINKED ISSUE #575: Add tensor slice operation
At the moment, taking a slice of a tensor is very difficult. The only ways to do it are to use the select operation (which is pretty cumbersome for this purpose) or to extract the data, do the slice manually, and then construct a new tensor.

My specific use case involves dividing a tensor into four quadrants, as a data augmentation strategy.

nkoppel:
Resolves #575 and #591.

Todo:

- [x] More tests
- [x] Documentation
- [x] More dtypes for `slice` and `reshape` (#591)


nkoppel:
My current implementation of adding dtypes to slice and reshape doesn't implement backward functions for the integer dtypes, so these will error at runtime if the backward operations are called for these dtypes on cuda.

coreylowman:
Awesome, nice work!

PR #603, opened 2023-03-23
[WIP] Upscale2D and ConvTrans2d

opfromthestart:
Adds the two image operations

To use upscale, add an `Upscale2D<OH, OW>` layer, which will upscale the input image to a `(B, C, Ow, OH)` from a `(B, C, W, H)`.


src/tensor_ops/upscale2d/mod.rs
```diff
@@ -0,0 +1,210 @@
...
+            }
+        }
+        impl<T> $TryTrait for T {}
+
+        impl<
+                C: Dim,
+                const H: usize,
+                const W: usize,
+                const OH: usize,
+                const OW: usize,
```
coreylowman, line 85: I'm actually thinking we don't need const dimensions for upscale. I think we can take the approach we have with [broadcast](https://github.com/coreylowman/dfdx/blob/main/src/tensor_ops/broadcast_to.rs#L17), where we do:

```rust
trait Upscale<E> {
    fn upscale2d<OH: ConstDim, OW: ConstDim>(self) -> ...;
    fn try_upscale2d<OH: ConstDim, OW: ConstDim>(self) -> ...;
    fn upscale2d_like<OH: Dim, OW: Dim>(self, size: (OH, OW)) -> ...;
    fn try_upscale2d_like<OH: Dim, OW: Dim>(self, size: (OH, OW)) -> ...;
}
```

src/tensor_ops/upscale2d/mod.rs
```diff
@@ -0,0 +1,210 @@
...
+        }
+    };
+}
+
+upscale2d!(
+    Kernel = NearestUpscale2DKernel,
+    ConstTrait = ConstNearestUpscale2D,
+    TryTrait = TryNearestUpscale2D,
+    Meth = nearest_upscale2d,
+    TryMeth = try_nearest_upscale2d
```
coreylowman, line 177: Should we support other options for how to upscale? Could accept an enum, but that'd be much more complex of an implementation

coreylowman:
I think we could probably separate these two into separate PRs if thats easier for you. And then adding nn layers for both can be separate PRs as well.

src/tensor_ops/upscale2d/mod.rs
opfromthestart: I think having an Upscale operation could be a trait, which is then implemented on Nearest and Bilinear, I'd like at least those two.

src/tensor_ops/convtrans2d/mod.rs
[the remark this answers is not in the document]
opfromthestart: It is very WIP, I'm pretty sure that the implementation is still straight from Conv

src/tensor_ops/upscale2d/mod.rs
opfromthestart: For the trait yeah, the module still will though.

opfromthestart:
It may be easier as multiple PRs but it isn't really usable until all of those are done, so I think it makes more sense to just have a large PR with all of it.

src/tensor_ops/convtrans2d/mod.rs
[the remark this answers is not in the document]
opfromthestart: Ok, it was actually much closer to the forward direction than I thought, all I had to do is switch the index calulating code from forward to backward. I have 2 tests written for it so far.

opfromthestart:
I think this is now officially in the tesing phase, all of the things that needed done are done.

coreylowman:
Awesome work, I'll check this out tomorrow!

src/nn/mod.rs
```diff
@@ -256,6 +259,7 @@ pub mod modules {
         TransformerEncoder, TransformerEncoderBlock,
     };
     pub use super::unbiased_linear::UnbiasedLinear;
+    pub use super::upscale::Upscale2D;
```
coreylowman, line 262: You'll need to add to the builders mod below as well

src/nn/upscale.rs
```diff
@@ -0,0 +1,112 @@
...
+    type Error = Img::Err;
+
+    fn try_forward(&self, x: Img) -> Result<Self::Output, Img::Err> {
+        x.try_upscale2d()
+    }
+}
+
+#[cfg(feature = "nightly")]
+#[derive(Debug, Default, Clone)]
+pub struct Upscale2DBy<const H: usize, const W: usize = H, M: UpscaleMethod = NearestNeighbor>(M);
```
coreylowman, line 27: Nice!

src/tensor_ops/convtrans2d/cuda_kernel.rs
```diff
@@ -0,0 +1,180 @@
...
+        let params = (op, lhs.data.as_ref(), &img_strides, &mut patches);
+        unsafe { unfold_fn.launch(cfg, params) }?;
+
+        // (O, C * K * K) * (B, C * K * K, OH * OW) = (B, O, OH * OW)
+        let m = op.chan_out;
+        let k = op.chan_in * op.kernel * op.kernel;
+        let n = op.h_out * op.w_out;
+        unsafe {
+            sgemm_batch(
+                self.blas.as_ref(),
```
coreylowman, line 79: This was recently changed (sorry about that!). Can you merge main into this branch?
```suggestion
            self.gemm_batch(
```

src/tensor_ops/upscale2d/upscale2d.cu
```diff
@@ -0,0 +1,243 @@
...
+    T *out // 4d (Batch, Channels, HeightOut, WidthOut)
+) {
+    unsigned int i = blockIdx.x * blockDim.x + threadIdx.x;
+    const size_t numel = op.batch * op.chan * op.h_out * op.w_out;
+    if (i >= numel) {
+        return;
+    }
+
+    float h_scale = ((float)(inp_strides[1]/inp_strides[2]))/(out_strides[1]/out_strides[2]);
+    float w_scale = ((float)(inp_strides[2]/inp_strides[3]))/(out_strides[2]/out_strides[3]);
```
coreylowman, line 27: Have you tried this with broadcasted tensors? Should this be using the shape instead of strides?

src/tensor_ops/upscale2d/upscale2d.cu
```diff
@@ -0,0 +1,243 @@
...
+    idx /= op.batch;
+
+    // Probably isn't efficient, but it works
+    size_t oh_s = 0;
+    size_t ow_s = 0;
+    size_t oh_e = op.h_out-1;
+    size_t ow_e = op.w_out-1;
+    while (oh_s*h_scale < y) {
+        oh_s++;
+    }
```
coreylowman, line 83: Is this `size_t oh_s = y / h_scale + 1`?

src/nn/upscale.rs
```diff
@@ -0,0 +1,112 @@
...
+}
+
+#[cfg(feature = "nightly")]
+impl<
+        const H: usize,
+        const W: usize,
+        const IH: usize,
+        const IW: usize,
+        const B: usize,
+        const C: usize,
```
coreylowman, line 67: We can probably make these generic dims
```suggestion
        B: Dim,
        C: Dim,
```

src/nn/upscale.rs
```diff
@@ -0,0 +1,112 @@
...
+#[cfg(feature = "nightly")]
+impl<const H: usize, const W: usize, M: UpscaleMethod> NonMutableModule for Upscale2DBy<H, W, M> {}
+
+#[cfg(feature = "nightly")]
+impl<
+        const H: usize,
+        const W: usize,
+        const IH: usize,
+        const IW: usize,
+        const C: usize,
```
coreylowman, line 40: Can probably be a generic dim
```suggestion
        C: Dim,
```

src/tensor_ops/mod.rs
```diff
@@ -252,9 +252,17 @@ pub(crate) use to_dtype::ToDtypeKernel;
...
+
+#[cfg(feature = "nightly")]
+mod convtrans2d;
 #[cfg(feature = "nightly")]
-pub(crate) use conv2d::TryConv2DTo;
+pub use convtrans2d::{ConvTransAlgebra, TryConvTrans2D, TryConvTrans2DTo};
+
+mod upscale2d;
+pub use upscale2d::{
+    Bilinear, ConstUpscale2D, NearestNeighbor, TryUpscale2D, Upscale2DKernel, UpscaleMethod,
```
coreylowman, line 264: Don't want to export kernels 😁 
```suggestion
    Bilinear, ConstUpscale2D, NearestNeighbor, TryUpscale2D, UpscaleMethod,
```

src/tensor_ops/convtrans2d/mod.rs
```diff
@@ -0,0 +1,584 @@
...
+    /// w = torch.rand((2, 3, 3, 3), requires_grad=True)
+    /// print(x)
+    /// print(torch.swapaxes(w, 0, 1))
+    /// y = torch.conv_transpose2d(x, w)
+    /// print(y)
+    /// y.exp().mean().backward()
+    /// print(x.grad)
+    /// print(torch.swapaxes(w.grad, 0, 1))
+    /// ```
+    fn convtrans2d_test() {
```
coreylowman, line 244: any thoughts on more easily writing tests & increasing code coverage for complex operations like conv2d & conv transpose? They are some of the more complicated ones to test (along with transformers), and the current method we have for testing leaves a lot to be desired

coreylowman (changes requested):
This is looking great! Really coming along. Just a few changes that I think would help. Let me know if you want me to help and I can throw some PRs into your branch

src/tensor_ops/upscale2d/upscale2d.cu
opfromthestart: I dont think so if h_scale divides y.

src/nn/mod.rs
opfromthestart: I did this for ConvTrans2D, but neither upscale has a builder.

src/tensor_ops/convtrans2d/mod.rs
opfromthestart: I think as long as the tests have a tensor with all dimensions of input and output >1 and both the output and grads match with pytorch, I think its fine. Were you thinking more along the lines of tests that don't rely on pytorch/ are randomly generated?

src/tensor_ops/upscale2d/upscale2d.cu
opfromthestart: It should be yes.

opfromthestart: This should be fixed.

src/tensor_ops/convtrans2d/mod.rs
coreylowman: Yeah exactly, and in a way where we can test larger tensors. Mostly thinking about testing conv2d in cuda. The kernels are fairly complex, and some issues might not be caught unless we have fairly large sizes on things.

coreylowman:
@opfromthestart looks like this contains the changes for leakyrelu/prelu, can you remove those so they stay separate?

src/nn/mod.rs
coreylowman: Ah yeah, but currently builders re-exports modules if there's no existing builder for it. The intent is someone can either do `use dfdx::modules::*` or `dfdx::builders::*` and get everything either way.

coreylowman (approved):
This looks good to me - once the prelu and leakyrelu are removed i'll merge! Will probably be a bunch of other updates to apply to this separately to this pr including:
1. Updating conv transpose kernels & launching based on recent updates to conv2d ones
2. Making both conv transpose & upscale compatible with runtime dimensions

coreylowman:
Awesome work @opfromthestart!! Super great contribution 🚀

PR #615, opened 2023-03-25
Fix no-std support

### LINKED ISSUE #612: Add cargo check with no-std to check workflow
Currently there are no checks to ensure no_std compatability.

The command is this: `cargo check --all --no-default-features -F no-std`, which can be as a step to the cargo-check.yml workflow

### LINKED ISSUE #613: Fix no-std support
Related to #612 

A few recent PRs are not no_std compatible. You can see the issues by running `cargo check --all --no-default-features -F no-std`.

This issue should go through and fix the issues found with the command above. It should mainly be imports

Alexandcoats:
# Description

Fixes `no-std` support and adds CI check to avoid future breakages.

EDIT: I added [`cargo hack`](https://github.com/taiki-e/cargo-hack) to the CI to test feature combos more robustly than simply checking `no-std` by itself, and I removed the error that is created when trying to use both `std` and `no-std` together, opting instead to simply override `no-std` in that case.

## Related Issues

- Closes #612
- Closes #613

### DIFF
--- a/src/tensor/storage_traits.rs
+++ b/src/tensor/storage_traits.rs
@@ -430,15 +430,7 @@ pub trait TensorFrom<Src, S: Shape, E: Unit>: DeviceStorage {
 
 impl<E: Unit, D: DeviceStorage + TensorFromVec<E>> TensorFrom<E, Rank0, E> for D {
     fn try_tensor(&self, src: E) -> Result<Tensor<Rank0, E, Self>, Self::Err> {
-        #[cfg(feature = "no-std")]
-        let buf = {
-            let mut buf = Vec::with_capacity(1);
-            buf.push(src);
-            buf
-        };
-        #[cfg(not(feature = "no-std"))]
-        let buf = vec![src];
-        self.try_tensor_from_vec(buf, ())
+        self.try_tensor_from_vec(vec![src], ())
     }
 }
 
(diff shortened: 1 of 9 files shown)


src/tensor/storage_traits.rs
```diff
@@ -430,15 +430,7 @@ pub trait TensorFrom<Src, S: Shape, E: Unit>: DeviceStorage {
...
-        #[cfg(feature = "no-std")]
-        let buf = {
-            let mut buf = Vec::with_capacity(1);
-            buf.push(src);
-            buf
-        };
-        #[cfg(not(feature = "no-std"))]
-        let buf = vec![src];
-        self.try_tensor_from_vec(buf, ())
+        self.try_tensor_from_vec(vec![src], ())
```
coreylowman, line 433: I think vec! is only present with std, The old version should work with both?

src/tensor_ops/slice/cpu_kernel.rs
```diff
@@ -18,8 +18,6 @@ impl<E: Unit> SliceKernel<E> for Cpu {
             .get_strided_index(inp.shape.first_idx_in_slice(slice));
         let view = &inp.data[start_idx..];
 
-        println!("{} {}", start_idx, inp.shape.first_idx_in_slice(slice));
```
coreylowman, line 21: Nice 😀

coreylowman:
The MKL & cuda features aren't gonna work on ci unfortunately. Is there a way to blacklist features with cargo hack?

src/tensor/storage_traits.rs
Alexandcoats: The macro import from `alloc` seems to cover this.

coreylowman: Sweet, thanks!

coreylowman (approved):
Thanks for this, super helpful!

PR #616, opened 2023-03-25
Adds Tensor::recip, sigmoid/square/sqrt use existing kernels

coreylowman:


src/tensor_ops/sigmoid.rs
```diff
@@ -21,20 +15,20 @@ pub struct SigmoidKernelOp;
...
-impl<S: Shape, E: Dtype, D: UnaryKernel<SigmoidKernelOp, E>, T: Tape<E, D>> Tensor<S, E, D, T> {
+impl<S: Shape, E: Dtype, D: Device<E>, T: Tape<E, D>> Tensor<S, E, D, T> {
     /// See [sigmoid]
     pub fn sigmoid(self) -> Self {
         self.try_sigmoid().unwrap()
     }
     /// See [sigmoid]
     pub fn try_sigmoid(self) -> Result<Self, D::Err> {
-        try_unary_op(SigmoidKernelOp, self)
+        self.try_negate()?.try_exp()?.try_add(E::ONE)?.try_recip()
```
nkoppel, line 31: Is calling 4 kernels for an activation function a good idea? This is bound to be a little slower, and will use 4 times more memory during training,

nkoppel:
The relu implementation can also be factored out with `tensor.max(E::default())'

src/tensor_ops/sigmoid.rs
coreylowman: Without operator fusion probably not a good idea - I think it could actually be much much slower since it'll have to alloc 4 times instead of 1 times.

Issue #553, opened 2023-03-11
nvcc does not recognize gpu_architecture option `native`


quietlychris:
When compiling `dfdx` using `features = ["nightly", "cuda"]`, I run into an issue compiling the CUDA kernels using nvcc. In particular, I receive the following error for all kernels.
```
nvcc fatal   : Value 'native' is not defined for option 'gpu-architecture'
```

Per [this](https://github.com/jytime/Mask_RCNN_Pytorch/issues/15) issue, I ran `nvcc --list-gpu-code` which resulted in 
```
sm_35
sm_37
sm_50
sm_52
sm_53
sm_60
sm_61
sm_62
sm_70
sm_72
sm_75
sm_80
sm_86
sm_87
```
I swapped out the `native` option in the `build.rs` file for `sm_87` (an arbitrary choice for the highest number, I'm a little bit out of my wheelhouse here), which then seemed to compile fine. Additionally, the `nvcc --help` section under `--gpu-architecture` says
```
Allowed values for this option:  'all','all-major','compute_35','compute_37',
        'compute_50','compute_52','compute_53','compute_60','compute_61','compute_62',
        'compute_70','compute_72','compute_75','compute_80','compute_86','compute_87',
        'lto_35','lto_37','lto_50','lto_52','lto_53','lto_60','lto_61','lto_62',
        'lto_70','lto_72','lto_75','lto_80','lto_86','lto_87','sm_35','sm_37','sm_50',
        'sm_52','sm_53','sm_60','sm_61','sm_62','sm_70','sm_72','sm_75','sm_80',
        'sm_86','sm_87'.
```

My `nvcc --version` information is 
```
nvcc: NVIDIA (R) Cuda compiler driver
Copyright (c) 2005-2021 NVIDIA Corporation
Built on Thu_Nov_18_09:45:30_PST_2021
Cuda compilation tools, release 11.5, V11.5.119
Build cuda_11.5.r11.5/compiler.30672275_0
```
on a 4070 Ti running on Pop!_OS 22.04 with a 6.0.12 kernel.

coreylowman:
I guess native was added after 11.5? If we can figure out what the minimum version is for native we can fallback to all in cases where native doesn't exist.

coreylowman:
Related to #438

PR #619, opened 2023-03-25
Switch to using nvcc --list-gpu-code for build.rs compute_cap

quietlychris:
This is a notional change for `build.rs` for addressing what still seems to be a build issue relating to #553 (as `dfdx` is not currently compiling on main for me), switching to using the `nvcc` list of acceptable GPU codes instead. 

My `nvcc --version` info is
```
Cuda compilation tools, release 11.5, V11.5.119
Build cuda_11.5.r11.5/compiler.30672275_0
```
and `nvidia-smi` info of 
```
NVIDIA-SMI 525.89.02    Driver Version: 525.89.02    CUDA Version: 12.0
```
with the hardware the same as in the original issue. 

I'm more than happy to add in changes that might make this compatible with other development environments, in case this doesn't work with other people's.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -65,15 +65,23 @@ mod cuda {
             let start = std::time::Instant::now();
 
             let compute_cap = {
-                let out = std::process::Command::new("nvidia-smi")
-                    .arg("--query-gpu=compute_cap")
-                    .arg("--format=csv")
+                let out = std::process::Command::new("nvcc")
+                    .arg("--list-gpu-code")
                     .output()
                     .unwrap();
                 let out = std::str::from_utf8(&out.stdout).unwrap();
-                let mut lines = out.lines();
-                assert_eq!(lines.next().unwrap(), "compute_cap");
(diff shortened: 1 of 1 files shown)


coreylowman:
Huh weird, thanks for opening this. Can you post the output of the two commands? I'm curious what's different between them.

```
$ nvidia-smi --query-gpu=compute_cap --format=csv
$ nvcc --list-gpu-code
```

quietlychris:
Sure thing--the results are below 
```sh 
$ nvidia-smi --query-gpu=compute_cap --format=csv
compute_cap
8.9
```
and
```sh
$ nvcc --list-gpu-code
sm_35
sm_37
sm_50
sm_52
sm_53
sm_60
sm_61
sm_62
sm_70
sm_72
sm_75
sm_80
sm_86
sm_87
```

coreylowman:
Interesting I'm surprised they are different. Thanks for this fix then!

PR #620, opened 2023-03-26
Allow conv2d and pool2d to use dynamic dimensions for width and height.

nkoppel:
Adds the `convolve_dim` method to ConvAlgebra and uses it to implement ConvAlgebra for usize. Conv2d and Pool2d still require nightly, as making them work on stable would require making `Const<N>::Convolved` be `usize`, which I don't think is a desirable outcome of changing toolchains.

### DIFF
--- a/src/tensor_ops/conv2d/mod.rs
+++ b/src/tensor_ops/conv2d/mod.rs
@@ -78,8 +78,10 @@ pub(super) trait Conv2DKernel<E: Dtype>: DeviceStorage {
     ) -> Result<(), Self::Err>;
 }
 
-pub trait ConvAlgebra<const K: usize, const S: usize, const P: usize>: ConstDim {
-    type Convolved: ConstDim;
+pub trait ConvAlgebra<const K: usize, const S: usize, const P: usize>: Dim {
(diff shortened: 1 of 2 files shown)


src/tensor_ops/conv2d/mod.rs
```diff
@@ -88,6 +90,18 @@ where
...
+    fn convolve_dim(&self) -> Self::Convolved {
+        Self::Convolved::default()
+    }
+}
+
+impl<const K: usize, const S: usize, const P: usize> ConvAlgebra<K, S, P> for usize {
+    type Convolved = usize;
+
+    fn convolve_dim(&self) -> Self::Convolved {
+        (self.size() + 2 * P - K) / S + 1
```
coreylowman, line 103:
```suggestion
        (self.size() + 2 * P).checked_sub(K).unwrap() / S + 1
```

I think we'll need to add some sort of check/assertion for when this conv operation is valid. The const generic version does this implicitly with the where bound

coreylowman:
I like how this ends up removing more code than it adds😀

coreylowman:
Nice change 🚀

PR #621, opened 2023-03-26
Implement concat for usize arrays; add concat to Device.

nkoppel:
Small QOL improvements to the concat operation.


coreylowman:
👍

PR #622, opened 2023-03-26
Fix bug in reshape on cuda

nkoppel:
Fixes cuda implementation of reshape not allocating enough memory for broadcasted tensors and adds test to verify the new behavior. Also fixes 02-ops not compiling on cuda, although I don't know when/how this bug was introduced.


coreylowman (approved):
Ty for this one - Broadcasting is such a surprising pain, would like to go through every tensor op and add broadcast tests for it

PR #624, opened 2023-03-27
Revert "Switch to using nvcc --list-gpu-code for build.rs compute_cap…

coreylowman:
… (#619)"

This reverts commit 67f48840ce05a742e5b9e4ef155cde17148bd33e.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -65,23 +65,15 @@ mod cuda {
             let start = std::time::Instant::now();
 
             let compute_cap = {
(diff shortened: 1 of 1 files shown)


coreylowman:
@quietlychris can you upgrade your nvcc? Since your nvidia-smi has a version above what nvcc can compile to, maybe updating would fix this? If so, we can update build.rs to output an error message if it detects differences in nvidia-smi and nvcc

PR #626, opened 2023-03-28
[WIP] JIT conv2d kernel

coreylowman:
Related to #578 

Optimizng conv kernels and also moving to JIT compiling it instead of at compile time.


src/tensor_ops/conv2d/cuda_kernel.rs
```diff
@@ -68,8 +63,31 @@ where
...
+            let src = src.replace("op.kernel", &op.kernel.to_string());
+            let src = src.replace("op.batch", &op.batch.to_string());
+            let src = src.replace("op.chan_in", &op.chan_in.to_string());
+            let src = src.replace("op.chan_out", &op.chan_out.to_string());
+            let src = src.replace("op.h_in", &op.h_in.to_string());
+            let src = src.replace("op.h_out", &op.h_out.to_string());
+            let src = src.replace("op.w_in", &op.w_in.to_string());
+            let src = src.replace("op.w_out", &op.w_out.to_string());
+            let src = src.replace("$TY", Self::TYPENAME);
+            let ptx = compile_ptx(src).unwrap();
```
coreylowman, line 80: @nkoppel thoughts on this? This makes the first pass through the network take a bit longer since it has to compile the PTX, but:
1. Its less kernels to compile at compile time
2. we can inline values

I'm just fooling around with this for now (it didn't make that much of a performance impact for conv2d), but curious what others might think of JITing kernels

nkoppel: This seems like a pretty good idea, especially considering that conv2d is the kind of operation that is likely to be run several times with the exact same parameters. However, I think that we should generally avoid making shapes a part of kernels' compile time configuration, so that we can guarantee a finite number of kernels are created.

I would also be interested to know if there is benefit to caching kernels in files over recompiling them every run.

coreylowman:
I'm going to close this for now, but i'll keep the branch. I'll open an issue where we can discuss what ops to JIT compile

PR #632, opened 2023-03-29
Add ability to include smaller last batch

nkoppel:
This is a breaking change that changes the name of `batch` to `batch_exact`, and redefines `batch` to include the smaller last batch if it exists. As documented, the last batch will always contain at least one element.


coreylowman (approved):
Nice changes! 👍

Issue #580, opened 2023-03-17
Support for 4 bit Quantization


vikigenius:
Language model progress has been rapid recently and with the llama weights being released, so much progress is being made on the c++ side

https://github.com/ggerganov/llama.cpp

I see that fp16 is on the roadmap soon.

But it might also be a good idea to consider support for 4 bit quantization and related techniques. Is that something that will be considered?

coreylowman:
Does anyone know how they represent tensors with 4 bit data? Would this be some packed structure where they store 2 i4s with a u8?

```rust
struct i4x2(u8)
```

nkoppel:
[Here](https://github.com/ggerganov/ggml/pull/27) is a relevant description of the implementation llama.cpp uses to represent floats in 4 bits. It essentially boils down to storing some number of 4 bit integers along with a f32 scaling factor and an optional f32 offset. From what I have read of source code, it seems possible to do a lot of the math highly efficiently on the cpu using SIMD packed 8/16 bit integers, and without touching floating point at all.

While this representation is excellent for specialized inference libraries, I don't think that it's practical for a generalist library like dfdx, because dfdx must deal with strided indexing and must support cuda. Furthermore, I'm not sure how efficient we could make operations on this representation, considering that simd support in stable rust is very limited.

vikigenius:
Yep, I was also looking into this, this would be very nice to have, but I am not sure if we can make it even remotely approach the efficiency of what the C++ people are doing, considering how general purpose dfdx is.

I think the SIMD concerns are not that big of a deal, since dfdx heavily relies on nightly for some features already anyway and the SIMD support there is ok from what I have seen so far.

But I take your point on how specialized this is, and I am not sure if it is worth the effort to have this representation as an option for dfdx tensors.

PR #633, opened 2023-03-29
feat: add quantized CPU device

Alexandcoats:
# Description

This PR is an initial attempt at supporting 4-bit quantized tensors. There is likely still very much wrong with this, but it technically works (sort of).

- Adds a `QuantizedCpu` which uses a `type Storage = QuantizedStorage<impl Quantize>`
- Refactors `DeviceStorage` into several traits, and moves the generic `Unit` type to the trait level to support finer bounds
- Adds Unary/Binary kernels for `QuantizedCpu` (unfinished)

## Open Questions

- Because each block of values shares a single set of quantization factors (`scaling_factor`, `offset_factor`), the values must be recomputed whenever any other value changes. For this reason I implemented the kernels by iterating first over blocks, then over values, then recomputing the block (on drop). For this reason I did not allow arbitrary mutable access to individual values in the tensor. Perhaps there is a technique to avoid that that I am unaware of.

- The `BLOCK_SIZE` constant should probably be on the `QuantizedCpu` and inherited by the `QuantBlock` but this introduces the necessity of the `where [(); BLOCK_SIZE / 2]: Sized` bound across most of the surrounding code. It's awful.

- SIMD???

- Big tensors fail for various reasons. I have no idea why.

## Related Issues

- #580


coreylowman:
Nice idea to add a separate device for this! I bet we could even do something like `struct Quantized<D: DeviceStorage>(D)`

> Adds a QuantizedCpu which uses a type Storage = QuantizedStorage<impl Quantize>

Am I correct that you had to move away from GATs because there was no way to add `E: Quantize` to the dtype when implementing for QuantizedCPU? ie.
```rust
impl DeviceStorage for QuantizedCpu {
    type Storage<E: Unit> = Vec<u8>;
}
```
wouldn't necessarily work because we don't know that you can quantize E? E.g. if E is a bool

Alexandcoats:
Yes, exactly. It was simply not possible to specify that trait impl without doing that.

coreylowman:
Also thoughts on having some `Quantized<E: Quantize + Unit>(u4)` type that we use for Dtype instead of having different device? Then we could still implement specific kernels for `Quantized<f32>`

src/tensor/cpu/device.rs
```diff
@@ -60,12 +60,8 @@ impl HasErr for Cpu {
...
 }
 
-impl DeviceStorage for Cpu {
-    type Vec<E: Unit> = Vec<E>;
-
-    fn try_alloc_grad<E: Unit>(&self, other: &Self::Vec<E>) -> Result<Self::Vec<E>, Self::Err> {
-        self.try_alloc_zeros(other.len())
-    }
+impl<E: Unit> DeviceStorage<E> for Cpu {
+    type Storage = Vec<E>;
```
coreylowman, line 64: I wonder if it's worthwhile to make this non-generic. I'm imagining:
```rust
impl DeviceStorage<f32> for Cpu { type Storage = Vec<f32> }
impl DeviceStorage<f64> for Cpu { type Storage = Vec<f64> }
impl DeviceStorage<bool> for Cpu { type Storage = Vec<bool> }
impl DeviceStorage<usize> for Cpu { type Storage = Vec<usize> }
...
impl DeviceStorage<Quantized<f32>> for Cpu { type Storage = Vec<QuantBlock> }
```

coreylowman: & then we wouldn't need a separate QuantizedCpu device

coreylowman: but we'd likely have to make the kernel impls non-generic too. Because I don't *think* you can have?
```rust
impl<E: Float> UnaryKernel<AbsKernelOp, E> for Cpu { }
impl UnaryKernel<AbsKernelOp, Quantized<f32>> for Cpu  { }
```

Alexandcoats: Either way we'll need completely different impls for the kernels, iters, etc. No way around that unless we can match the access pattern of a `Vec`.

coreylowman: Yep totally agree there, I like the approach of having `DeviceStorage<E>` - it fits well, and we gain a lot by this method. For instance we could represent sparse tensors by
```rust
struct Sparse<E>(E);

struct MaskedVec<E> {
    data: Vec<E>,
    mask: Vec<bool>,
}

impl<E> DeviceStorage<Sparse<E>> for Cpu {
    type Storage = MaskedVec<E>
}
```

which isn't possible with the current GAT approach.

Plus `Tensor<S, Sparse<E>, D>` or `Tensor<S, Quantized<E>, D>` looks pretty dang cool!

Alexandcoats:
I'm closing this because I don't think there's enough of a use-case to justify the work necessary to implement it.

Issue #630, opened 2023-03-28
JIT compile cuda kernels


coreylowman:
As shown in #626, with the next cudarc release, it will be possible to JIT compile cuda kernels.

_Originally posted by @coreylowman in https://github.com/coreylowman/dfdx/issues/626#issuecomment-1487608378_

Blocking questions for this:
1. What kernels should we JIT compile?
2. What, if any, values should we inline into the kernel while compiling? Possible options may be:
    a. Operation parameters (e.g. for conv2d we could inline stride/kernel/padding)
    b. Num elements
    c. num dimensions
    d. shape
    e. strides

nkoppel:
I've already posted some of my stance on this in #626, but I'll restate and clarify it here. 

In general, it should be fine to inline anything that is known at compile-time, but most runtime values should be passed in, assuming JIT compilations take at least 10 ms. This is because inlining variables that take on several values during execution would cause a significant slowdown. For example, shape and stride should not be included because users may run their networks on a large variety of input sizes, when could occur when running a cnn on unscaled and uncropped images from the internet.

As long as we keep the number of compilations small, there shouldn't be much downside to JIT compiling every operation, except for slow startup.

coreylowman:
We could do different things in training vs inference - since training will have very regular inputs, however I think it'd be too complex internally to have a feature flag to swap between jit vs ahead of time IMO. So not inlining shapes/strides makes sense to me

coreylowman:
I think the biggest wins here will be with kernels that need to work with many different types. Also kernels that are more niche that are less likely to be used so we don't have to compile them unless they are used

PR #634, opened 2023-03-30
JIT compiling kernel for to_dtype & reshape

coreylowman:
Bumps cudarc version & enables nvrtc feature to gain access to `compile_ptx` for JIT compiling kernels.

The following kernels are now JIT compiled:
1. to_dtype
2. reshape

Related to #630


src/tensor_ops/reshape_to/mod.rs
```diff
@@ -73,7 +75,29 @@ pub trait ReshapeTo: HasErr + HasShape {
...
+                    tape: self.tape,
+                })
+            } else {
+                self.device.forward(dst, &self)
+            }
+        })
+    }
+}
+
+impl<S: Shape, E: Dtype, D: ReshapeKernel<E>> ReshapeTo for Tensor<S, E, D, OwnedTape<E, D>> {
```
coreylowman, line 100: Hmm apparently this doesn't work because anything using a generic tape `T` would need to ensure that `Tensor<..., T>: ReshapeTo`

nkoppel: The only thing I can think of to fix this is using a macro to generate implementations for everything but bool, which has a backward implementation of `unimplemented!()` or something

coreylowman: Going to revert this specific change - I addressed this in #635

src/tensor_ops/reshape_to/cuda_kernel.rs
```diff
@@ -102,3 +83,97 @@ where
         Ok(())
     }
 }
+
+const FWD_KERNEL: &str = "
+#if __WORDSIZE == 64
+typedef long int intptr_t;
+#else
+typedef int intptr_t;
+#endif
```
coreylowman, line 92: Noting that you can't use `<cstdin>` or `"stdint.h"` in nvrtc because of include paths.

According to https://forums.developer.nvidia.com/t/nvrtc-missing-stdint-h/146318, including standard headers like this should be avoided. The only reason we need the above is for intptr_t, so its easy to redefine them as the above.

PR #648, opened 2023-03-31
adding usize dtype to cuda_kernel

zojeda:
Very small contribution... but it is a start :)

### DIFF
--- a/src/tensor_ops/stack/stack.cu
+++ b/src/tensor_ops/stack/stack.cu
@@ -24,8 +24,8 @@ extern "C" __global__ void sum_f64(
 
 extern "C" __global__ void sum_usize(
     const size_t numel,
-    const unsigned int *inp,
-    double *out
+    const size_t *inp,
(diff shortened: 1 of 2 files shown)


src/tensor_ops/stack/stack.cu
```diff
@@ -21,3 +21,15 @@ extern "C" __global__ void sum_f64(
     }
     out[i] += inp[i];
 }
+
+extern "C" __global__ void sum_usize(
+    const size_t numel,
+    const unsigned int *inp,
+    double *out
```
coreylowman, line 28:
```suggestion
    const size_t *inp,
    size_t *out
```

coreylowman (changes requested):
Thanks for opening this! Pretty easy one to review 😁 You'll need to update the types of the sum_usize function, other than that this looks good though!

coreylowman (approved):
Thanks for the contribution!

PR #651, opened 2023-04-01
Adds cudnn feature flag. Removes "test-cuda" feature flag. Using cuDNN for convolutions

### LINKED ISSUE #527: Use the winograd convolution algorithm when possible
See https://medium.com/@dmangla3/understanding-winograd-fast-convolution-a75458744ff for beginner details.

https://arxiv.org/abs/1509.09308 is the paper
> Deep convolutional neural networks take GPU days of compute time to train on large data sets. Pedestrian detection for self driving cars requires very low latency. Image recognition for mobile phones is constrained by limited processing resources. The success of convolutional neural networks in these situations is limited by how fast we can compute them. Conventional FFT based convolution is fast for large filters, but state of the art convolutional neural networks use small, 3x3 filters. We introduce a new class of fast algorithms for convolutional neural networks using Winograd's minimal filtering algorithms. The algorithms compute minimal complexity convolution over small tiles, which makes them fast with small filters and small batch sizes. We benchmark a GPU implementation of our algorithm with the VGG network and show state of the art throughput at batch sizes from 1 to 64. 

There are a ton of articles on this!

### LINKED ISSUE #638: Use cuDNN for conv2d
Now that Cuda has a workspace method, it's possible to use cuDNN to speed up convolutions. The work for this should be:

1. Add a `cudnn` feature to dfdx `cudnn = ["cuda", "cudarc/cudnn"]`
2. In conv kernel forward & backward, if cudnn feature is active, invoke the cudnn API calls, otherwise keep the existing methods.

Note that currently cudarc doesn't have a safe cudnn API, so this may have to use straight sys/result calls for now. Part of this issue can be contributing back to cudarc as well.

coreylowman:
Resolves #638
Resolves #527 

TODOS
- Should we optionally enable cudnn?

Timings of `cargo +nightly bench --bench conv2d -F cuda` on an A10

| branch | fwd | bwd |
| --- | --- | --- |
| main | 4.73ms | 20ms |
| this | 2.6ms | 8ms |

### DIFF
--- a/.github/workflows/cargo-check-features.yml
+++ b/.github/workflows/cargo-check-features.yml
@@ -9,9 +9,9 @@ jobs:
       matrix:
         config:
           - toolchain: stable
-            command: cargo hack check --feature-powerset --no-dev-deps --depth 2 --skip default,nightly,cpu-mkl-matmul,cuda,test-cuda
+            command: cargo hack check --feature-powerset --no-dev-deps --depth 2 --skip default,nightly,cpu-mkl-matmul,cuda,cudnn
           - toolchain: nightly
-            command: cargo hack check --each-feature --no-dev-deps --features nightly --skip default,cpu-mkl-matmul,cuda,test-cuda
+            command: cargo hack check --each-feature --no-dev-deps --features nightly --skip default,cpu-mkl-matmul,cuda,cudnn
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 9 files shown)


opfromthestart:
How does this compare with the old implementation on the conv2d benchmark?

coreylowman:
@opfromthestart updated PR description with timings, its much faster 😀

PR #658, opened 2023-04-03
Fixes conv transpose stride bug, adds more docs to upscale2d

### LINKED ISSUE #657: Odd behavior when training
I am making a CNN based autoencoder, and it will not train properly with large batches. If I feed it a large batch (16), the error will not decrease, or will increase over time. However, when given batches of 1, the error does decrease. I am using some of the new/experimental features (Conv2D, ConvTrans2d, Upscale, PReLU) so I suspect that there is an error in one of them that doesn't calculate gradients right for batched inputs.
The repo of my code is at https://github.com/opfromthestart/touhou-ai, but it probably only runs on my computer at the moment. The structure of the model is at https://github.com/opfromthestart/touhou-ai/blob/master/src/net/dfdx.rs#L13.

coreylowman:
Resolves #657 

There were two bugs here:
1. patches stride was incorrect in the actual kernel - was using op.chan_in instead of op.chan_out
2. In the backward pass gemm_batched call, it was not broadcasted the filters (so using m * n instead of 0)

Additionally I added a bunch of tests and cleaned up the implementations.

I also changed upscale methods to accept the method as a paramter. This is for a number of reasons:
1. This allows us to actually add args to each of the methods in the future, and use them in the methods.
2. It unifies the compile & run time shape calls.


src/tensor_ops/upscale2d/mod.rs
```diff
@@ -74,53 +81,83 @@ pub trait Upscale2DWithMethod<M: UpscaleMethod>: HasErr {
...
+/// let t: Tensor<Rank3<3, 32, 32>, f32, _> = dev.zeros();
+/// let y: Tensor<Rank3<3, 64, 64>, f32, _> = t.upscale2d(NearestNeighbor);
+/// ```
+///
+/// Runtime upscale:
+/// ```rust
+/// # use dfdx::prelude::*;
+/// # let dev: Cpu = Default::default();
+/// let t: Tensor<Rank3<3, 32, 32>, f32, _> = dev.zeros();
+/// let y: Tensor<(Const<3>, usize, usize), f32, _> = t.upscale2d_like(NearestNeighbor, 64, 64);
```
coreylowman, line 115: The interface to upscale2d has changed. It now takes in the method as a parameter so you can call it like in these examples (notably without specifying any generics in either case). This unifies how you call them in both const/runtime situations, as before you would've had to specify the method in the generic both ways

PR #659, opened 2023-04-03
Some QOL fixes

opfromthestart:
Specify dimensions in Upscale error when trying to upscale to a smaller size.
Puts PReLU in its own file to make importing easier.


src/nn/prelu.rs
```diff
@@ -0,0 +1,205 @@
...
+pub struct PReLU<E: Dtype, D: Device<E>> {
+    pub a: Tensor<(), E, D>,
+}
+
+impl<E: Dtype, D: Device<E>> NonMutableModule for PReLU<E, D> {}
+
+impl<S: ConstShape, E: Dtype, D: Device<E>, T: Tape<E, D>> Module<Tensor<S, E, D, T>>
+    for PReLU<E, D>
+where
+    Tensor<S, E, D, T>: TryPReLU<Tensor<S, E, D, NoneTape>>,
```
coreylowman, line 43: You actually don't need these anymore! 😀

PR #666, opened 2023-04-05
Reshape layer

opfromthestart:
Layer for reshaping. Currently crashes at runtime if not equal, but I would like for it to be able to do that.


opfromthestart:
And something else, how do I get my fork to not have all these extra commits?

coreylowman (approved):
Clean PR 😀 👍

PR #668, opened 2023-04-05
Adding cudnn kernel for sum

coreylowman:
Adding a cudnn kernel for sum forward. There's no corresponding backward operation for reductions in cudnn, so it just re-uses the cuda backward method.


coreylowman:
@nkoppel cudnn is slower in the repo benchmarks than your sum kernel. nice job! (it also doesn't handle reducing pre-broadcasted tensors, so not sure how to support those...)

nkoppel:
Interesting! Might have something to do with cudnn's need for a workspace or the ordering of the output tensor.

coreylowman:
Closing as this doesn't make improvements

PR #676, opened 2023-04-08
Have SplitInto keep tapes of each head seperate

nkoppel:
Changes the behavior of SplitInto to propagate the tape of the input to the output of the first head, and to keep the operations of each head in its respective output.


coreylowman:
Resolves #676

coreylowman:
Awesome, nice change 👍

PR #680, opened 2023-04-08
Simplify upscale cuda kernels

### LINKED ISSUE #677: Upscale2D or Conv2D produces NaN values
My application using dfdx: https://github.com/nkoppel/generals-bot is currently failing to train a unet to play generals.io because the [policy network](https://github.com/nkoppel/generals-bot/blob/072bb025c4ce67b5e5756fa0878bcb4f8677c8e5/src/bots/ppgnet.rs#L80) produces NaN values during training. This is probably a bug in Conv2D or Upscale2D, as the NaN values only occupy a small area of the output, and this program has successfully trained networks without Upscale2D. The NaN values occur about every 100-500 batches, and appear only in the last image of a batch, in the following shape:

```
0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0
0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0
0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0
0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0
0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0
0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0
0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0
0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0
0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 NaN
0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 NaN
NaN NaN NaN NaN NaN NaN NaN NaN NaN NaN
``` 
(along the bottom and the bottom 3 of the right edge, for all 4 output channels)

Due to its relative rarity and the consistency of the shape of the NaN values, I suspect that this is a bug due to Conv2D or Upscale2D reading uninitialized memory or failing to write to uninitialized memory, but I have been unable to locate the bug thus far. It is probably more likely to be a bug in Upscale2D, because I have tried using the new cudnn implementation of Conv2D, and experienced the exact same behavior.

This error is reproduced here: https://gist.github.com/nkoppel/db07f684222a7169d2c39b205542d4c5. I have no idea why, but using Sgd or inspecting the value of `out` both prevent the bug from occurring.

coreylowman:
Resolves #677

### DIFF
--- a/src/tensor_ops/upscale2d/upscale2d.cu
+++ b/src/tensor_ops/upscale2d/upscale2d.cu
@@ -33,8 +33,8 @@ __device__ void nearest_upscale2d_fwd(
     idx /= op.chan;
     const size_t b = idx % op.batch;
 
-    size_t ih = min(static_cast<size_t>(h_scale * oh), op.h_out - 1);
-    size_t iw = min(static_cast<size_t>(w_scale * ow), op.w_out - 1);
+    size_t ih = min(static_cast<size_t>(h_scale * oh), op.h_in - 1);
+    size_t iw = min(static_cast<size_t>(w_scale * ow), op.w_in - 1);
 
     size_t inp_i = b * inp_strides[0] + c * inp_strides[1] + ih * inp_strides[2] + iw * inp_strides[3];
     
@@ -65,8 +65,8 @@ __device__ void nearest_upscale2d_bwd(
     idx /= op.chan;
     const size_t b = idx % op.batch;
 
-    size_t ih = min(static_cast<size_t>(h_scale * oh), op.h_out - 1);
-    size_t iw = min(static_cast<size_t>(w_scale * ow), op.w_out - 1);
+    size_t ih = min(static_cast<size_t>(h_scale * oh), op.h_in - 1);
+    size_t iw = min(static_cast<size_t>(w_scale * ow), op.w_in - 1);
 
     size_t inp_i = b * inp_strides[0] + c * inp_strides[1] + ih * inp_strides[2] + iw * inp_strides[3];
     atomicAdd(grad_inp + inp_i, grad_out[i]);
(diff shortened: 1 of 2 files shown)


coreylowman:
@nkoppel for some reason this seems to resolve the bug? I have no idea why. On main I get an illegal memory address accessed. But not on this branch

nkoppel:
This seems to fix it on my end as well, though I am somewhat skeptical that it's not due to a different allocation pattern.

src/tensor_ops/upscale2d/upscale2d.cu
```diff
@@ -98,12 +95,11 @@ __device__ void bilinear_upscale2d_fwd(
...
-    idx /= op.batch;
 
-    size_t y0 = min(static_cast<size_t>(h_scale * oh), op.h_out - 1);
-    size_t y1 = min(y0 + 1, op.h_out - 1);
-    size_t x0 = min(static_cast<size_t>(w_scale * ow), op.w_out - 1);
-    size_t x1 = min(x0 + 1, op.w_out - 1);
+    size_t y0 = min(static_cast<size_t>(h_scale * oh), op.h_in - 1);
+    size_t y1 = min(y0 + 1, op.h_in - 1);
+    size_t x0 = min(static_cast<size_t>(w_scale * ow), op.w_in - 1);
+    size_t x1 = min(x0 + 1, op.w_in - 1);
```
coreylowman, line 102: @nkoppel this was the issue

Issue #683, opened 2023-04-09
Check that `nvcc` can compile to compute cap in build.rs


coreylowman:
Throw helpful error message if not

quietlychris:
I was just looking into this, actually (sorry that I've been offline with respect to the question you asked in #624, I haven't had a lot of spare capacity for open source recently).

Unfortunately, it looks like I'm unable to update `nvcc` in order to get that matching; I've double-checked that I'm using the latest drivers that are available through System76 packaging (see [this](https://support.system76.com/articles/cuda/) and [this](https://support.system76.com/articles/system76-driver/#installing-the-system76-nvidia-driver-for-systems-with-nvidia-gpus)). While I might be able to do a manual install, I'd like to avoid this if possible, and I think that it would be beneficial from a community perspective to support this type of system configuration; Pop!_OS is commonly used in the single-machine deep learning community due to System76's built-in CUDA management, and it's relatively common due to its popularity within the Linux gaming community, all of whom are potential users of `dfdx`. 

I was thinking that maybe we could try a nested solution? Throw an error if the current check in `build.rs` doesn't work, which can then be caught and try to use the code I submitted in #619, before otherwise erroring out? I know it's not the most elegant solution, but I think it could cover quite a few potential set-ups.

coreylowman:
Thanks for the update. I think nested would work well. Sounds like the logic would be:

1. Grab compute_cap from nvidia-smi
2. Grab available gpu codes from nvcc (using your code)
3. If nvidia-smi compute_cap is higher than the highest gpu code from nvcc (your case), then choose the highest gpu code in nvcc
4. Otherwise, nvidia-smi compute_cap should be present in nvcc gpu codes, and we can use that

The case that I had was nvidia-smi compute_cap was less than highest nvcc gpu code, so in the new logic we'd just choose nvidia-smi since it's present in nvcc gpu codes.

coreylowman:
Thanks for the PR @quietlychris, works great on my end!

PR #685, opened 2023-04-09
Initial merging of nvidia-smi and nvcc checks

quietlychris:
This addresses the discussion in #683. I've checked that this compiles with `features=nightly,cuda` on my machine, but if this doesn't work for you or I misunderstood what you had in mind, just let me know!

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -68,7 +68,8 @@ mod cuda {
         {
             let start = std::time::Instant::now();
 
-            let compute_cap = {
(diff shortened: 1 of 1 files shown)


coreylowman (approved):
Yeah this looks good to me - I'll probably also add a check for compute_cap in the nvcc codes and panic with a helpful message if not. That should be very unlikely, but good to cover all the bases.

PR #688, opened 2023-04-10
feat: use `Cow` when appropriate

Alexandcoats:
# Description

Some tensor ops use `Result<Borrowed, Owned>` when a `Cow` would be more idiomatic. This fixes those instances.

### DIFF
--- a/src/tensor_ops/pow/cuda_kernel.rs
+++ b/src/tensor_ops/pow/cuda_kernel.rs
@@ -24,7 +24,7 @@ where
     fn forward<S: Shape>(
         &self,
         op: super::PowiKernelOp,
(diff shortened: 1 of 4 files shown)


src/tensor_ops/pow/cuda_kernel.rs
```diff
@@ -24,7 +25,7 @@ where
     fn forward<S: Shape>(
         &self,
         op: super::PowiKernelOp,
-        inp: Result<&Tensor<S, E, Self>, Tensor<S, E, Self>>,
+        inp: Cow<Tensor<S, E, Self>>,
```
coreylowman, line 28: This is awesome, nice 🔥

PR #689, opened 2023-04-10
feat: add `Tensorlike` to clean up spooky ghosts

Alexandcoats:
# Description

Some tensor ops allow using either the Tensor with the `Arc` data reference or a `GhostTensor` without it. Currently it uses `Result` to do this, but this is a bit awkward. I found that a unifying trait `Tensorlike` is enough to handle all of the current scenarios and is cleaner.

### DIFF
--- a/src/tensor/tensorlike.rs
+++ b/src/tensor/tensorlike.rs
@@ -1,0 +1,65 @@
+use crate::{
+    prelude::{HasErr, HasShape},
+    shapes::{Shape, Unit},
+    tensor::DeviceStorage,
+};
+
+use super::{storage_traits::AllocGrad, GhostTensor, Tensor, UniqueId};
+
+/// Contains everything that comprises a tensor, except possibly for the actual data. This really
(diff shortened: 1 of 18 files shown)


src/tensor/tensorlike.rs
```diff
@@ -0,0 +1,65 @@
...
+    shapes::{Shape, Unit},
+    tensor::DeviceStorage,
+};
+
+use super::{storage_traits::AllocGrad, GhostTensor, Tensor, UniqueId};
+
+/// Contains everything that comprises a tensor, except possibly for the actual data. This really
+/// exists to unify handling of [Tensor] and [GhostTensor].
+///
+/// *If it looks like a tensor and barks like a tensor, then pet it like a tensor.*
```
coreylowman, line 12: 🤣

coreylowman (approved):
I like it, using impl TensorLike is much clearer than the result with tensor/GhostTensor, nice! Mind fixing the conflicts & merging main branch in?

coreylowman:
Nice, thanks for the improvement 👍

PR #690, opened 2023-04-11
Add `contiguous` and `try_contiguous` methods

### LINKED ISSUE #679: Add `contiguous()`/`try_contiguous()` as a method to `trait ReshapeTo`
These two methods can just have default implementations of calling reshape:

```rust
fn try_contiguous(self) -> Result<Self, D::Err> {
   self.try_reshape_like(self.shape()) 
}
```

This works because reshape already ensures that output is contiguous, and calling reshape with the same shape of the tensor will ensure it has the same shape.

VasanthakumarV:
Resolves #679,

Used the methodology specified in #679, if I missed something please do let me know.

### DIFF
--- a/src/tensor/ghost.rs
+++ b/src/tensor/ghost.rs
@@ -29,8 +29,29 @@ impl<S: Shape, E: Unit, D: DeviceStorage, T> Tensor<S, E, D, T> {
     }
 }
 
+impl<S: Shape, E: Unit, D: DeviceStorage> Clone for GhostTensor<S, E, D> {
+    fn clone(&self) -> Self {
+        Self {
+            id: self.id,
+            len: self.len,
+            shape: self.shape,
+            strides: self.strides,
+            dev: self.dev.clone(),
+            marker: self.marker,
+        }
+    }
+}
+
 impl<S: Shape, E: Unit, D: DeviceStorage> super::storage_traits::HasErr for GhostTensor<S, E, D> {
     type Err = D::Err;
+}
+
+impl<S: Shape, E: Unit, D: DeviceStorage> HasShape for GhostTensor<S, E, D> {
+    type WithShape<New: Shape> = GhostTensor<New, E, D>;
+    type Shape = S;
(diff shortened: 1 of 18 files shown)


src/tensor_ops/reshape_to/mod.rs
```diff
@@ -68,6 +68,23 @@ pub trait ReshapeTo: HasErr + HasShape {
...
+    fn contiguous(self) -> Self::WithShape<Self::Shape>
+    where
+        Self::Shape: ConstShape,
+    {
+        self.try_contiguous().unwrap()
+    }
+    /// See [`ReshapeTo::contiguous`]
+    fn try_contiguous(self) -> Result<Self::WithShape<Self::Shape>, Self::Err>
+    where
+        Self::Shape: ConstShape,
```
coreylowman, line 83: We actually don't need these bounds, mind removing it from both of the methods?
```suggestion
    {
        self.try_contiguous().unwrap()
    }
    /// See [`ReshapeTo::contiguous`]
    fn try_contiguous(self) -> Result<Self::WithShape<Self::Shape>, Self::Err>
```

src/tensor_ops/reshape_to/mod.rs
```diff
@@ -212,4 +229,17 @@ mod tests {
...
+    fn test_contiguous() {
+        let dev: TestDevice = Default::default();
+
+        let a: Tensor<_, TestDtype, _> = dev.tensor([[0.1, 0.2, 0.3], [0.4, 0.5, 0.6]]);
+
+        let b1 = a.clone().contiguous();
+        assert_eq!(a.strides, b1.strides);
+
+        let b2: Tensor<_, TestDtype, _> = a.permute::<Rank2<3, 2>, _>().contiguous();
+        assert_eq!(b2.strides, [2, 1]);
```
coreylowman, line 243: Nice unit test 👍

coreylowman (changes requested):
Looks really good, will merge once we remove the extra bounds

PR #698, opened 2023-04-12
Add examples of runtime dimensions to `examples/02-ops.rs`

### LINKED ISSUE #593: Add more examples of runtime dimensions to examples/02-ops.rs
Currently examples/01-tensor.rs shows how to create tensors with runtime dimensions, but there are no corresponding examples in 02-ops.rs.

1. Add some of the tensors in examples/02-tensor to use both runtime & const dimensions.
2. Change the first matmul call to use both runtime & const dims
3. Show an example of using `RealizeTo` trait to add tensors with different types of dimensions:
```rust
let a: Tensor<(Const<3>, usize), f32, _> = dev.sample_uniform_like(&(Const, 5));
let b: Tensor<(usize, Const<5>), f32, _> = dev.sample_uniform_like(&(3, Const));
let c = a + b.realize().unwrap();
```

VasanthakumarV:
Resolves #593,

I have mostly followed the instructions in #593, the only deviation is introducing a `realize` example in `01-tensor.rs`, thought this will be a prerequisite to understand the new examples in `02-ops.rs`.

### DIFF
--- a/examples/01-tensor.rs
+++ b/examples/01-tensor.rs
@@ -3,6 +3,7 @@
 use dfdx::{
     shapes::{Const, Rank1, Rank2, Rank3},
     tensor::{AsArray, OnesTensor, SampleTensor, Tensor, TensorFrom, ZerosTensor},
+    tensor_ops::RealizeTo,
 };
 
 #[cfg(not(feature = "cuda"))]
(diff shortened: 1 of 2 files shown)


coreylowman (approved):
Nice 👍 I like the addition of realize!

PR #707, opened 2023-04-14
Optimize chunk reduction algorithms

nkoppel:
Optimizes chunk_sum, chunk_min, and chunk_max by using the [__shfl_down_sync](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#warp-shuffle-functions) intrinsic to transfer memory between threads instead of shared memory. Despite causing more contention in the output buffer, this implementation measures ~25% faster on the `sum` benchmark, and I was unable to reduce contention in a way that measurably improves performance.

### DIFF
--- a/src/tensor_ops/max_to/max_to.cu
+++ b/src/tensor_ops/max_to/max_to.cu
@@ -40,13 +40,12 @@ __device__ void chunk_max(
     unsigned int chunk_start = max((int)(warp_i - chunk_i), 0);
     unsigned int chunk_end = min((unsigned int)(warp_i + chunk_len - chunk_i), warpSize);
 
-    unsigned int mask = (1 << chunk_end) - (1 << chunk_start);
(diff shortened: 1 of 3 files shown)


coreylowman:
Thanks for the PR! What GPU are you using for benchmarking? On an A10 there's not a difference on sum for me (for batchnorm/softmax is slightly slower actually).

Also any thoughts on doing block level reductions based on this article? https://developer.nvidia.com/blog/faster-parallel-reductions-kepler/

nkoppel:
Interesting, newer gpu architectures must have faster shared memory. My gpu is a T500, an older laptop gpu.

coreylowman:
For future reference: https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/cuda/Reduce.cuh

nkoppel:
Closing this for now, may reopen if I think of a way to apply pytorch's implementation.

PR #709, opened 2023-04-16
Improve buffer caching

nkoppel:
So far, makes the TensorCache api cleaner using the CacheStorage trait to allow calling buffer conversion logic within the context of a TensorCache.

## Todo:
- [ ] More thoroughly document safety details
- [ ] Add buffer removal strategy (FIFO until total size of the cache is less than some number of bytes)
- [ ]  Set a reasonable default maximum cache size, and allow users to configure cache size.

### DIFF
--- a/src/tensor/cache.rs
+++ b/src/tensor/cache.rs
@@ -33,12 +33,37 @@ pub(crate) struct AllocationKey {
 /// valid allocation. When the last value is removed from the list, the key
 /// is removed.
 #[derive(Debug)]
-pub(crate) struct TensorCache<Ptr> {
+pub(crate) struct TensorCache<Ptr: CacheStorage> {
     pub(crate) allocations: RwLock<BTreeMap<AllocationKey, Vec<Ptr>>>,
     pub(crate) enabled: RwLock<bool>,
 }
 
-impl<Ptr> Default for TensorCache<Ptr> {
+pub(crate) trait CacheStorage: Sized {
+    type Output<T>: CacheStorage;
+
+    /// Unsafely converts the elements of a contiguous collection type to another type. Note:
+    /// **This function is wildly unsafe**, see implementations for details
+    unsafe fn transmute_elements<T>(self) -> Self::Output<T>;
+
+    /// Uses transmute_elements to convert to an element type with alignment `align` before dropping. 
+    unsafe fn drop_with_alignment(self, align: usize) {
+        match align {
+            1 => drop(self.transmute_elements::<u8>()),
+            2 => drop(self.transmute_elements::<u16>()),
+            4 => drop(self.transmute_elements::<u32>()),
+            8 => drop(self.transmute_elements::<u64>()),
+            16 => drop(self.transmute_elements::<u128>()),
+            _ => panic!("Invalid alignment")
+        }
+    }
+
+    /// calls [CacheStorage::drop_with_alignment], extracting `align` from an [AllocationKey]
+    unsafe fn drop_with_key(self, key: AllocationKey) {
+        self.drop_with_alignment(key.alignment)
+    }
+}
+
+impl<Ptr: CacheStorage> Default for TensorCache<Ptr> {
     fn default() -> Self {
         Self {
             allocations: Default::default(),
@@ -47,7 +72,7 @@ impl<Ptr> Default for TensorCache<Ptr> {
     }
 }
 
-impl<Ptr> TensorCache<Ptr> {
+impl<Ptr: CacheStorage> TensorCache<Ptr> {
     /// Returns the number of allocations in the cache.
     #[allow(unused)]
     pub(crate) fn len(&self) -> usize {
@@ -102,7 +127,7 @@ impl<Ptr> TensorCache<Ptr> {
 
     /// Returns a cached allocation if one exists.
     /// Otherwise, returns `None`.
-    pub(crate) fn try_pop<E>(&self, len: usize) -> Option<Ptr> {
+    pub(crate) fn try_pop<E>(&self, len: usize) -> Option<Ptr::Output<E>> {
         if !self.is_enabled() {
             return None;
         }
@@ -142,21 +167,23 @@ impl<Ptr> TensorCache<Ptr> {
             if items.is_empty() {
                 cache.remove(&key);
             }
-            Some(allocation)
+            Some(unsafe { allocation.transmute_elements() })
         } else {
             None
         }
     }
 
     /// Inserts an allocation into the cache.
-    pub(crate) fn insert<E>(&self, len: usize, allocation: Ptr) {
+    pub(crate) fn insert<E>(&self, len: usize, allocation: Ptr::Output<E>)
+        where Ptr::Output<E>: CacheStorage<Output<u8> = Ptr>
+    {
         if !self.is_enabled() {
-            // This is a panic because it's a bug in the library.
-            panic!("Tried to insert into a disabled cache.");
+            return;
         }
 
+        let allocation = unsafe { allocation.transmute_elements::<u8>() };
         let layout = Layout::new::<E>();
-        let num_bytes = len * std::mem::size_of::<E>();
+        let num_bytes = len * layout.size();
         let key = AllocationKey {
             num_bytes,
             size: layout.size(),
@@ -181,6 +208,21 @@ impl<Ptr> TensorCache<Ptr> {
             cache.get_mut(&key).unwrap().push(allocation);
         }
     }
+
+    pub(crate) fn clear(&self) {
+        #[cfg(not(feature = "no-std"))]
+        let mut cache = self.allocations.write().unwrap();
+        #[cfg(feature = "no-std")]
+        let mut cache = self.allocations.write();
+        for (&key, allocations) in cache.iter_mut() {
+            assert!(key.num_bytes % key.size == 0);
+            assert!(key.num_bytes < isize::MAX as usize);
+            for alloc in allocations.drain(..) {
+                unsafe { alloc.drop_with_key(key) };
+            }
+        }
+        cache.clear();
+    }
 }
 
 #[cfg(test)]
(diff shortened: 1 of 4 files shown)


build.rs
```diff
@@ -93,8 +93,8 @@ mod cuda {
                 let out = out.lines().collect::<Vec<&str>>();
                 let mut codes = Vec::with_capacity(out.len());
                 for code in out {
-                    let code = code.split("_").collect::<Vec<&str>>();
-                    if code.len() != 0 && code.contains(&"sm") {
+                    let code = code.split('_').collect::<Vec<&str>>();
+                    if !code.is_empty() && code.contains(&"sm") {
```
nkoppel, line 97: This just fixes clippy warnings.

src/tensor/cache.rs
```diff
@@ -29,80 +72,181 @@ pub(crate) struct AllocationKey {
...
+
+impl<Ptr: CacheStorage> Default for TensorCache<Ptr> {
     fn default() -> Self {
         Self {
             allocations: Default::default(),
             enabled: RwLock::new(false),
+            drop_queue: Default::default(),
+            size: RwLock::new(0),
+            // TODO: default max size
+            max_size: RwLock::new(1_000_000),
```
nkoppel, line 176: @coreylowman Thoughts on a default `max_size`? I think that 1gb is reasonable in most cases, but maybe we can do something like a percentage of available system memory or we could require the user to specify this value upon enabling the cache.

nkoppel: I've decided to require users to specify the maximum size of the cache when they enable it.

coreylowman:
So are these the main benefits of this PR?
1. Enable maximum cache size
2. More unified transmutation/drop logic across devices

nkoppel:
Yes, and this should also be more memory safe.

src/tensor/cache.rs
```diff
@@ -1,24 +1,70 @@
...
     /// The alignment of the allocation in bytes - from [Layout].
-    pub alignment: usize,
+    alignment: usize,
+}
+
+#[derive(Debug)]
+struct AllocationGroup<Ptr: CacheStorage> {
+    // Tracks the number of matching 'AllocationKey's in drop_queue to ignore. This is used to
+    // "remove" the next instance of the matching AllocationKey in the drop_queue, without having
+    // to run an O(n)
```
coreylowman, line 65: I think I'm missing something - an O(n) what? `vec.remove(0)` would have O(n) performance, but `vec.pop()` shouldn't do any moving of the data.

src/tensor/cache.rs
```diff
@@ -1,24 +1,70 @@
...
+use alloc::vec;
 
 #[cfg(not(feature = "no-std"))]
 use std::sync::RwLock;
 
 #[cfg(feature = "no-std")]
 use spin::RwLock;
 
-/// A key for the tensor cache. Contains both number of bytes and informatino
+macro_rules! read {
```
coreylowman, line 19: 👍

coreylowman (commented):
I like the size limiting idea a lot, I don't even think that's configurable in pytorch so that'd be a great add!

Will need to comb over to determine if safety is still met. The new layers of abstraction here do make it a bit more difficult to reason about the safety at each part, which may be a downside of trying to abstract between cpu/cuda.

And something to think about: I think we may want to enable re-using allocations that are bigger than what is requested (e.g. I need 100MB, and the allocation cache gives me a 200MB buffer). These changes shouldn't make this harder or anything, but worth thinking about what changes would need to be made in both cases

src/tensor/cache.rs
coreylowman: Thoughts on using an enum to represent different options for this? I'm definitely going to forget what a plain usize would represent
```rust
dev.enable_cache(CacheSize::Unlimited)
dev.enable_cache(CacheSize::NumItems(1000))
dev.enable_cache(CacheSize::Bytes(10))
dev.enable_cache(CacheSize::MB(10))
dev.enable_cache(CacheSize::GB(10))
```

src/tensor/cache.rs
nkoppel: This should be better explained now.

nkoppel:
> Will need to comb over to determine if safety is still met. The new layers of abstraction here do make it a bit more difficult to reason about the safety at each part, which may be a downside of trying to abstract between cpu/cuda.

I still think this is good for memory safety because it reduces the surface area of memory unsafety to CacheWrapper and CacheStorage, rather than having cache operations using memory unsafe operations directly.

> And something to think about: I think we may want to enable re-using allocations that are bigger than what is requested (e.g. I need 100MB, and the allocation cache gives me a 200MB buffer). These changes shouldn't make this harder or anything, but worth thinking about what changes would need to be made in both cases

I was actually thinking about this, and implementing this could be super easy due to the use of a BTreeMap, because BTreeMap's make it fast to get the next largest key after the one you requested. My main concern is that some operations may rely on the physical size of the buffer to get physical_numel or the number of threads to launch, and we would need to implement a `Tensor::physical_numel` method to instead calculate this using shape and strides.

src/tensor/cache.rs
coreylowman: Optional num bytes? `Option<usize>`

src/tensor/cuda/device.rs
```diff
@@ -1,19 +1,57 @@
...
+    ///
+    /// # Safety
+    ///
+    /// Assuming that `self` is a valid [CudaSlice], the main safety concern is that the output
+    /// slice should be assumed to be uninitialized
+    unsafe fn transmute_elements<T2>(self) -> Self::Output<T2> {
+        let dev = self.device();
+
+        let src_layout = Layout::new::<T>().pad_to_align();
+        let dst_layout = Layout::new::<T2>().pad_to_align();
```
coreylowman, line 38: I'm not sure if it's okay to use these with cuda - I don't think they follow the same layout/alignment rules as rust does.

It seems like they are just used to get number of bytes. Maybe just use `std::mem::size_of::<>() * self.data.len()`?

src/tensor/cache.rs
```diff
@@ -29,80 +75,215 @@ pub(crate) struct AllocationKey {
...
+
+    /// Uses transmute_elements to convert to an element type with alignment `align` before dropping.
+    /// This **must** be a memory safe way to drop self, given the correct alignment
+    unsafe fn drop_with_alignment(self, align: usize) {
+        match align {
+            1 => drop(self.transmute_elements::<u8>()),
+            2 => drop(self.transmute_elements::<u16>()),
+            4 => drop(self.transmute_elements::<u32>()),
+            8 => drop(self.transmute_elements::<u64>()),
+            16 => drop(self.transmute_elements::<u128>()),
```
coreylowman, line 114: On my dev machine u128 is aligned to 8, is it 16 for you?

src/tensor/cache.rs
```diff
@@ -29,80 +75,215 @@ pub(crate) struct AllocationKey {
...
+    /// returns the allocations's size in bytes
+    fn size(&self) -> usize;
+
+    /// Unsafely converts the elements of a contiguous collection type to another type. Note:
+    /// **This function is wildly unsafe**, see implementations for details
+    unsafe fn transmute_elements<T>(self) -> Self::Output<T>;
+
+    /// Uses transmute_elements to convert to an element type with alignment `align` before dropping.
+    /// This **must** be a memory safe way to drop self, given the correct alignment
+    unsafe fn drop_with_alignment(self, align: usize) {
```
coreylowman, line 108: Thoughts on removing default impl of this? This implementation is the CPU one, and I'm not sure we should be doing all of this for cuda.

src/tensor/cache.rs
```diff
@@ -1,24 +1,70 @@
...
-    pub alignment: usize,
+    alignment: usize,
+}
+
+#[derive(Debug)]
+struct AllocationGroup<Ptr: CacheStorage> {
+    // Tracks the number of matching 'AllocationKey's in drop_queue to ignore. This is used to
+    // "remove" the next instance of the matching AllocationKey in the drop_queue, without having
+    // to run an O(n) operation to actually remove the key.
+    ignore_drops: usize,
```
coreylowman, line 66: Is all this just because we have the invariant of "if a key is in the tree, it must have > 0 values"? I wonder if it would be better to just remove that assumption, and just keep keys in the tree forever.

Also, have you noticed speedups when doing this? I'm wondering how much of an impact on speed the key removal actually has. My main concern is that this will be hard to understand/maintain whenever we revisit. I think the performance benefit would have to be pretty high for this to be worth it IMO.

Thoughts?

src/tensor/cache.rs
```diff
@@ -115,89 +296,80 @@ impl<Ptr> TensorCache<Ptr> {
...
+            e.insert(AllocationGroup {
+                allocations: vec![allocation],
+                ignore_drops: 0,
+            });
         } else {
-            cache.get_mut(&key).unwrap().push(allocation);
+            cache.get_mut(&key).unwrap().allocations.push(allocation);
         }
+        std::mem::drop(cache);
+        self.shrink();
```
coreylowman, line 349: Think its worth only calling this if necessary?

src/tensor/cuda/device.rs
nkoppel: I think that cuda formats memory in much the same way as the cpu, as we are able to `dtoh` and `htod` directly from one to another, any my understanding is that this is a byte-for-byte copy of the host array. I believe that we should keep this in case we wish to arrays with structs, as this is able to compute the padded size of a type as it would appear in an array.

coreylowman: Yeah byte to byte I think is fine, I'm just not sure about the sizes/alignment.

pad_to_align says:
> Creates a layout by rounding the size of this layout up to a multiple of the layout’s alignment.
Which makes it sound like it can modify the length?

As far as alignment it sounds like cuda is always aligned to at least 256, which is definitely not the case for rust:

> Any address of a variable residing in global memory or returned by one of the memory allocation routines from the driver or runtime API is always aligned to at least 256 bytes.

https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#device-memory-accesses

src/tensor/cache.rs
nkoppel: I think that this is correct for cuda, and while it may not be necessary, it doesn't have a significant performance impact as transmute_elements is very fast.

src/tensor/cache.rs
nkoppel: I've made a small modification to `shrink` that minimizes the number of RwLock operations done by shrink when no shrinking needs to occur.

PR #710, opened 2023-04-16
Typos in feature_flags.rs

mauvray:
Hello, I noticed two small typos in the documentation,

For some context, I am learning rust so I copy pasted the cpu line (17) and got a mistake when running `cargo build` so I corrected the mistake.


coreylowman:
Thanks for fixing! 👍

PR #730, opened 2023-04-24
[RFC] Adding `thread_cpu()` and functions to create tensors on cpu

coreylowman:
This adds a `thread_cpu()` function akin to rand's `thread_rng()` function. This let's you get access to a global copy of a Cpu device, meaning you don't have to keep constructing cpu's in whatever method you are using one in.

Additionally, this let's us add functions for all the standard ways to create tensors, instead of relying on device methods:

```rust
use dfdx::tensor::*;
let _ = zeros::<Rank2<2, 3>, f32>();
let _ = ones::<Rank3<1, 2, 3>, f64>();
let _ = tensor([[1.0, 2.0, 3.0], [4.0, 5.0, 6.0]]);
```

TODOS
- [ ] Can we do something similar for Cuda?
- [ ] Add docstrings
- [ ] Add more tests

### DIFF
--- a/src/tensor/global.rs
+++ b/src/tensor/global.rs
@@ -1,0 +1,52 @@
+use super::*;
+use crate::shapes::*;
+
+thread_local! {
+    static GLOBAL_CPU: Cpu = Cpu::default();
+}
+
+pub fn thread_cpu() -> Cpu {
+    GLOBAL_CPU.with(|cpu| cpu.clone())
+}
+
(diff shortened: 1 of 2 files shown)


coreylowman:
For cuda we could do something like:

```rust
fn thread_cuda(ordinal: usize) -> Cuda {
    ...
}
```

Which could also invoke bind_to_thread.

However this probably wouldn't be super useful because you'd have to specify ordinal everywhere.

coreylowman:
Going to close for now

PR #735, opened 2023-04-25
Work on compatibility

ViliamVadocz:
- A few unrelated fixes to optimizers that I needed to compile and run the code.
- Compatibility header with a few functions:
  - Pay special attention to `atomicCAS` since it's quite unsafe and complicated. It's almost 3 AM so I worry I might have messed it up.
  - I implemented `__hmax_nan` and `__hmin_nan` so that we can have the nice NaN propagation even on older GPUs
  - Someone should try where the compute capability boundaries for when functions are defined actually lie. For now I just guessed `700`.
- I added an example Windows include path so that you have some idea of what it looks like for me.

---

I still get runtime errors like this:
```rs
---- tensor_ops::reshape_to::tests::test_1d_reshape_non_contiguous stdout ----
thread 'tensor_ops::reshape_to::tests::test_1d_reshape_non_contiguous' panicked at 'called `Result::unwrap()` on an `Err` value: CompileError { nvrtc: NvrtcError(NVRTC_ERROR_COMPILATION), options: ["--include-path=/usr/include", "--include-path=C:\\Program Files\\NVIDIA GPU Computing Toolkit\\CUDA\\v12.0\\include", "--gpu-architecture=sm_61"], log: "default_program(46): error: no instance of overloaded function \"atomicAdd\" matches the argument list\n            argument types are: (__half *, const __half)\n\n1 error detected in the compilation of \"default_program\".\n" }', src\tensor_ops\reshape_to\cuda_kernel.rs:79:56
```
which are very strange.

---

To be honest I think this PR breaks more things than it fixes. 🥴
For reference, running `cargo test --features cuda,test-f16` I get: `245 passed; 112 failed;`

### DIFF
--- a/src/tensor_ops/utilities/compatibility.cuh
+++ b/src/tensor_ops/utilities/compatibility.cuh
@@ -29,15 +29,15 @@ __device__ unsigned short int atomicCAS(
 ) {
     // Read the two shorts that make up the 32 bit int at the aligned memory location.
     unsigned int* aligned_address = (unsigned int*) ((size_t) address & ~2);
-    unsigned int my_short = *address;
+    // unsigned int my_short = *address;
     unsigned int other_short = *(unsigned short int*) ((size_t) address ^ 2);
     // Replace my_short with value in the integer.
     const unsigned int value = val;
     const bool aligned = ((size_t) address & 2) == 0;
-    unsigned int new_whole = aligned ? (value << 16) | other_short : (other_short << 16) | value;
+    unsigned int new_whole = aligned ? (other_short << 16) | value : (value << 16) | other_short;
     
     unsigned int old = atomicCAS(aligned_address, (unsigned int) compare, new_whole);
-    return aligned ? old >> 16 : old & (0xffff);
+    return aligned ? old & (0xffff) : old >> 16;
 }
 
 __device__ __half atomicAdd(__half* address, __half val) {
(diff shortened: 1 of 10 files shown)


src/tensor_ops/utilities/compatibility.cuh
```diff
@@ -0,0 +1,54 @@
...
+}
+#endif
+
+
+// https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#atomicadd
+// > The 16-bit __half floating-point version of atomicAdd() is only supported by devices of compute capability 7.x and higher.
+// We assume that atomicCAS for shorts has the same compute capability restrictions.
+#if __CUDA_ARCH__ < 700
+
+// Inspired by https://github.com/torch/cutorch/blob/master/lib/THC/THCAtomics.cuh#L96-L119
```
coreylowman, line 23: It seems like this is doing atomicAdd with `unsigned int` instead of `unsigned short`? Should we be doing that as well? Then we could basically just copy the function from the link without having to redefine atomicCAS

coreylowman:
The runtime errors you are getting are from the JIT compiled kernels using nvrtc. They don't have the `tensor_ops/utilities` as an include directory (since I'm not sure if that path even exists in compiled binaries), so you can't use things in cuda_utils.cuh or your new compatibility header.

Issue #729, opened 2023-04-24
Get size of tensor


sirandreww:
With the introduction of runtime dimensions there should be a simple way to get the shape of a tensor.
I imagine this would be similar to:
https://pytorch.org/docs/stable/generated/torch.Tensor.size.html

```rust
let a: Tensor<(usize, usize), f32, _> = dev.zeros_like(&(2, 3));
assert_eq!(a.size(), (2, 3));
```

There is no need to create another function that gets the size of a specific dimension like in the Pytorch case.
If you guys agree that this feature is nice to have. Where should the code for such a feature be added? 
I would like to work on it please.

coreylowman:
This is possible via `a.shape()`, which is accessible through the `dfdx::shapes::HasShape` trait.

Want to add an example of this to some of the examples?

Issue #727, opened 2023-04-23
Model saving in struct requires different type.


sirandreww:
I'm trying to create a struct that holds a MLP inside like so:
```
type Mlp = (
    (Linear<10, 32>, ReLU),
    (Linear<10, 32>, ReLU),
    (Linear<10, 32>, Tanh),
);

pub struct Predictor {
   model: Mlp,
}

impl Predictor {
    pub fn new() -> Self {
        let dev = AutoDevice::default();
        let m = dev.build_module::<Mlp, f32>();
        Self { model: m }
    }
}
```
This does not compile because the function `new` is trying to assign m to model, and they are not the same type.
To fix this i had to create a different type for saving the model:
```
type MlpSaved = (
    (Linear<10, 32, f32, Cpu>, ReLU), 
    (Linear<10, 32, f32, Cpu>, ReLU),
    (Linear<10, 32, f32, Cpu>, Tanh)
);

pub struct Predictor {
   model: MlpSaved,
}

type Mlp = (
    (Linear<10, 32>, ReLU),
    (Linear<10, 32>, ReLU),
    (Linear<10, 32>, Tanh),
);

impl Predictor {
    pub fn new() -> Self {
        let dev = AutoDevice::default();
        let m = dev.build_module::<Mlp, f32>();
        Self { model: m }
    }
}
```

Would it be possible to make an automatic conversion or something of the sort?

opfromthestart:
A cleaner way to do this would be
```rust
type Mlp = (
    (Linear<10, 32>, ReLU),
    (Linear<10, 32>, ReLU),
    (Linear<10, 32>, Tanh),
);

pub struct Predictor {
   model: <Mlp as BuildOnDevice<AutoDevice, f32>>::Built,
}

impl Predictor {
    pub fn new() -> Self {
        let dev = AutoDevice::default();
        let m = dev.build_module::<Mlp, f32>();
        Self { model: m }
    }
}
```

PR #740, opened 2023-04-26
Adding example

sirandreww:


sirandreww:
this is meant to solve:
 Model saving in struct requires different type. #727 
 Get size of tensor #729

examples/13-housing-nn-in-struct.rs
```diff
@@ -0,0 +1,234 @@
...
+
+// ************************************************************************************************
+// Model definition
+// ************************************************************************************************
+
+type Mlp = (
+    (Linear<2, 32>, ReLU),
+    (Linear<32, 32>, ReLU),
+    (Linear<32, 1>, Sigmoid),
+);
```
coreylowman, line 33: Thoughts on adding a type alias for the built version?

```suggestion
type MlpStructure = (
    (Linear<2, 32>, ReLU),
    (Linear<32, 32>, ReLU),
    (Linear<32, 1>, Sigmoid),
);
type Mlp = <MlpStructure as BuildOnDevice<AutoDevice, f32>>::Built;
```

coreylowman (changes requested):
Looks awesome! Left a couple of suggestions, but other than that this looks good to me.

examples/13-housing-nn-in-struct.rs
```diff
@@ -0,0 +1,234 @@
...
+
+fn function_we_would_like_the_nn_to_mimic(
+    input: Tensor<(Const<2>,), f32, AutoDevice>,
+) -> Tensor<(Const<1>,), f32, AutoDevice> {
+    let distance_from_center: Tensor<(), f32, AutoDevice> = input.powi(2).sum().sqrt();
+    if distance_from_center.as_vec()[0] > 1.0 {
+        AutoDevice::default().tensor([1.0])
+    } else {
+        AutoDevice::default().tensor([0.0])
+    }
```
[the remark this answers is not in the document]
sirandreww: Brilliant!

examples/13-housing-nn-in-struct.rs
sirandreww: Can you explain the reasoning behind this? what is the logical difference between MLP and MLPStructure ?

coreylowman:
Thanks for the contribution!! 🔥

PR #751, opened 2023-04-28
Allow `broadcast_like` to accept tensors OR shapes

### LINKED ISSUE #746: Allow broadcast_like to accept tensors OR shapes
Right now it accepts a shape, but it'd be conveinent to submit *either* a shape or a tensor. This is already how things like zeros_like/ones_like/sample_like work.

The changes required for this are:
1. Change `Dst: Shape` to be `Dst: HasShape` https://github.com/coreylowman/dfdx/blob/main/src/tensor_ops/broadcast_to.rs#L42
    1. Change the other usages of `Dst` to be `Dst::Shape`
2. Do the same change but for try_broadcast_like definition
3. Update the impl definition of try_broadcast_like to match the new definition https://github.com/coreylowman/dfdx/blob/main/src/tensor_ops/broadcast_to.rs#L58
4. Call `*dst.shape()` to get the new shape, isntead of using `dst` directly in the implementation

All the existing tests should pass, additionally, you should add a test for passing in a dynamically shaped tensor to broadcast_like.

VasanthakumarV:
Fixes #746

### DIFF
--- a/src/tensor_ops/broadcast_to.rs
+++ b/src/tensor_ops/broadcast_to.rs
@@ -29,45 +29,46 @@ pub trait BroadcastTo: HasErr + HasShape {
     where
         Self::Shape: BroadcastShapeTo<Dst, Ax>,
     {
-        self.try_broadcast_like(&Default::default()).unwrap()
+        self.try_broadcast_like::<Dst, _>(&Default::default())
+            .unwrap()
     }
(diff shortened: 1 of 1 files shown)


coreylowman (approved):
Hey thanks for opening this, these changes look perfect! Will merge shortly 🚀

Issue #186, opened 2022-09-12
Add multi-threaded cpu device


coreylowman:
This would likely use something like rayon under the hood to run operations inside a threadpool

iacore:
matrix multiplication is already multi-threaded (limited to 4 threads).

using https://lib.rs/crates/ggblas would remove this limitation

PR #764, opened 2023-05-03
[WIP] Adds rayon support to CPU

coreylowman:
Adds `"threaded"` feature, which enables rayon on CPU for most tensor operations.

TODOS:
- [ ] Add to batched matmul
- [ ] Add feature flag description
- [ ] Add to conv2d?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,7 +49,7 @@ indicatif = "0.17.3"
 glob = { version = "0.3.1", optional = true }
 
 [features]
-default = ["std", "fast-alloc", "cpu-par-matmul", "parallel"]
+default = ["std", "fast-alloc", "cpu-par-matmul"]
 nightly = ["half?/use-intrinsics"]
 
 std = ["cudarc?/std", "matrixmultiply?/std", "rand_distr/std_math"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,7 @@ num-traits = { version = "0.2.15", default-features = false }
 safetensors = { version = "0.3", default-features = false, optional = true }
 memmap2 = { version = "0.5", default-features = false, optional = true }
 half = { git = "https://github.com/starkat99/half-rs.git", branch = "main", optional = true, features = ["num-traits", "rand_distr"] }
+rayon = { version = "1.7.0", optional = true }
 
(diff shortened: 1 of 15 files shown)


M1ngXU:
i really need this 😅 (no cuda gpu anymore)
anything i can help you with and do you have an approx. ETA?

M1ngXU:
i also have a laptop but with 8 cores; maybe we could add another feature to only parallize cpu-heavy stuff? so all those small speedups with lots of overhead will still be single-threaded (and faster)?

coreylowman:
The 30 core machine is a VM that I used for GPU development haha.

It's definitely possible, we'd need to figure out a good threshold for it.

I'm curious what performance differences you see - can you run the `cargo bench --bench softmax` with and without the `-F threaded` and see what the difference is?

M1ngXU:
ok sure, gotta set up everything first though :D

M1ngXU:
hmm, threaded is about twice as slow (too) (500ms - 600ms vs 1300ms - 1500ms)

coreylowman:
I'll have to look into rayon a bit more I think. Maybe we should be setting up our own threadpool instead of using the global one?

coreylowman:
Oh, maybe we should be iterating over chunks of data (going off of comments in this thread https://www.reddit.com/r/rust/comments/5snkx4/using_rayon_to_parallelize_a_loop_makes_my/)?

coreylowman:
What chunk size are you using? 8 seemed better on my machine, but still not as good as non-threaded.

I'm still trying to figure out what exactly the overhead rayon introduces is. If we know the overhead then we can try to reduce it

M1ngXU:
i am using `(len() - 1) / std::thread::available_parallelism().unwrap().get() + 1`, so basically `len / available_parallelism` rounded up

M1ngXU:
and i've seen that we actually call extra methods on the threaded impl (didn't check which ones). maybe they have overhead too?

M1ngXU:
actually rust-analyzer always makes me go to the std::Iterator implementation if I follow the `for_each()` method (even though this should go to rayon) ...

coreylowman:
Gonna close this since not actively working on it. May restart work on it in the future. Also linking to #186

PR #769, opened 2023-05-07
Add LogSoftmax

kurnevsky:
It's useful for AlphaGo algorithm which uses LogSoftmax in its loss function.


kurnevsky:
It doesn't look like the error in pipeline is related to my changes? :)

coreylowman:
Yeah it definitely doesn't seem like it - which is weird because I haven't seen this issue on other workflows. I'll pull the changes down locally today and verify manually. Might be a quirk of nightly

coreylowman (approved):
Nice improvement!

Issue #770, opened 2023-05-07
Conv2D non-squared kernels/strides/padding


M1ngXU:
While this is usually not necessary, it would be nice (and I probably need it now 😅) to have non-squared kernel sizes, e.g. a height of 5 and a width of only 1.

M1ngXU:
i could try to impl it for cpu, probably also cudnn (though i don't have a nvidia card rn) and depending on the impl also cuda 😅

M1ngXU:
maybe have a type alias from `Conv2D` to `Conv2DGeneral` or so that specifies squared sizes

coreylowman:
Related to #392

M1ngXU:
> Related to #392

how didn't i see this issue 😅

M1ngXU:
close this one?

coreylowman:
Sure

PR #772, opened 2023-05-08
Conv2d general

M1ngXU:
#770

### DIFF
--- a/benches/conv2d.rs
+++ b/benches/conv2d.rs
@@ -12,7 +12,7 @@ fn main() {
     #[cfg(not(feature = "cuda"))]
     type Dev = Cpu;
 
-    type Model = Conv2D<128, 256, 4>;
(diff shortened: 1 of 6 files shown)


M1ngXU:
only for cpu rn, i'll try to impl cudnn (and maybe cuda)

M1ngXU:
cudnn should be easy, no idea about cuda though. also, how should i test it? i currently only test with the old tests, though this won't show errors if i someplace swapped `x` and `y`.

coreylowman:
Fyi @M1ngXU I just merged some pretty breaking changes for conv that you'll want to start with. It should be easier to work with hopefully

M1ngXU:
i'll `drop` this pr for now and just use it locally; maybe someone else can implement this. This should probably add non-squared stuff for pool/upsample/... too.

Issue #485, opened 2023-02-24
Non-tensor attributes of nn objects are not saved by SaveToNpz/LoadFromNpz


coreylowman:
This is currently affecting:
- BatchNorm2D::epsilon
- BatchNorm2D::momentum
- Dropout::p
- LayerNorm1D::epsilon

TimerErTim:
Would it be feasable to store those values as scalar tensor instead and let the `RecursiveWalker` handle the rest upon saving and loading?

coreylowman:
Yeah good idea, very straightforward!

PR #794, opened 2023-05-30
Mobilenet integration

### LINKED ISSUE #773: Add integration test for mobile net
With #767 now added, we can integration test against more complex image networks!

This new test should mirror the existing resnet18 integration test, but with a mobile net structure.

Awpteamoose:
#793 is prerequsite for this, closes #773

There's a very ugly hack with `fixup_batchnorms` that won't be necessary once #485 is resolved. There's also some type-level vodoo in `block.rs` to keep the actual model definition looking nice, but I could also just split the 5 kinds of blocks each into their own type.

For `SqueezeAndExcite` blocks my original idea was that I'd be able to just use `Linear` because as far as I understand, a 2d conv over a 1x1xN volume is just a linear operation, but maybe I need to transform the weights while saving or maybe there's something I don't understand.


Awpteamoose:
I can rebase on master and remove the fixup_batchnorms hack if this is still relevant.

PR #799, opened 2023-06-28
Add hyperparameter support to TensorCollection

### LINKED ISSUE #485: Non-tensor attributes of nn objects are not saved by SaveToNpz/LoadFromNpz

nkoppel:
Currently, adds infrastructure to use scalar values with the TensorCollection system. Resolves #485. A limitation of this is that boolean attributes cannot be used, as `bool` does not implement `num_traits::ToPrimitive`, and this cannot be worked around with conversions due to lifetime issues.

### DIFF
--- a/src/nn/tensor_collection/collection.rs
+++ b/src/nn/tensor_collection/collection.rs
@@ -6,7 +6,7 @@ use crate::{
     tensor_ops::Device,
 };
 
-use super::{ModuleField, ModuleFields, TensorField};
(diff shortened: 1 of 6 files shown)


coreylowman (approved):
Nice updates, thanks for this 🚀

Issue #755, opened 2023-04-29
Dynamic dimensions in neural network layers?


bplevin36:
The new dynamic dimension support for Tensors is great!  Are there any plans to support it for `Module`s as well?  To allow for e.g. a `Linear` layer where the input size is a const generic but the output size is dynamic?

I tried hacking this together myself, but ran into problems with allocation.  It seems like the existing `Device` APIs are pretty strongly angled towards being able to allocate a `Module` based purely on its type.  I can probably implement allocation and `Module` on my type manually, but it would be great to have a more ergonomic workaround.

coreylowman:
Glad you like it! Yeah aligning the existing `nn` layer to support both compile time & run time dimensions would complicate things quite a bit. It's definitely possible, for one example we could have linear defined as:
```rust
struct Linear<In: Dim, Out: Dim, E: Dtype, D: DeviceStorage> {
    weight: Tensor<(In, Out), E, D>,
    bias: Tensor<(Out, ), E, D>,
}
```

and then you could create type aliases like this:
```rust
type ConstLinear<const I: usize, const O: usize, E, D> = Linear<Const<I>, Const<O>, E, D>;
type DynLinear<E, D> = Linear<usize, usize, E, D>;
```

As far as instantiation, you're right the existing method assumes sizes known at compile time. We'd like have to do something similar to what tensors do where we have a version of creation with compile time shapes (`dev.zeros()`, and a separate version for runtime sizes (`dev.zeros_like(&(3, 5))`).

I think my main concern would be the increased complexity in adding these, both from internals perspective and also external usability perspective.

The other part of this is that currently in the deep learning ecosystem, most training/inference libraries redefine neural network types in their own libraries. huggingface does this all over this place. As far as dfdx goes, hopefully people can define their own `nn` layers similar to that. So if someone wanted to implement linear/transformer/etc with runtime shapes outside of dfdx, that should be possible.

Thoughts?

PR #803, opened 2023-07-03
[WIP] Adds `#[derive(Sequential)]` on nn builder structs

coreylowman:
Currently only tuples are supported for sequential networks. They are convenient and easy to write for small networks, but have a number of weaknesses when trying to scale up to larger networks.

1. The types get extremely long, making debug/error messages extremely difficult to read.
2. They are hard to work with when trying to accessing fields/subfields.

This PR introduces a new api for specifying sequential networks using a derive attribute:
```rust
#[derive(Sequential)]
struct MyMLP {
   layer1: Linear<5, 10>,
   act1: ReLU,
   layer2: Linear<10, 5>,
   act2: Tanh,
}
```

Internally, this will generate a new type that looks like this:

```rust
struct MyMLPBuilt<E: Dtype, D: Device<E>> {
   layer1: Linear<5, 10, E, D>,
   act1: ReLU,
   layer2: Linear<10, 5, E, D>,
   act2: Tanh,
}
```

and add all the TensorCollection/Module/ModuleMut implementations necessary.

This should be much easier to work with!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,3 +1,5 @@
+[workspace]
+
 [package]
 name = "dfdx"
 version = "0.11.2"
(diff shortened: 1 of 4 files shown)


coreylowman:
Noting that this effort has been moved into a separate repo https://github.com/coreylowman/slimnn. It's turned into a much bigger rewrite for the nn layer that will add:
1. `#[derive(Sequential)]` as mentioned in pr description
2. Simplify traits for nn layer (& remove TensorCollection)
3. Add derive macros for other new traits (ZeroGrads/ResetParams/UpdateParams/etc)
4. Add dynamically sized layers (#755).

I'm doing this work separately just for ease of implementation, and to see what it's like to implement a nn layer outside of the repo. Once that separate implementation is fairly complete, it will be merged into dfdx.

PR #811, opened 2023-07-12
[Breaking] Adds `AMP<F>` dtype

### LINKED ISSUE #424: Add AMP dtype
This depends on #423.

This would let each operation/optimizer choose which precision it uses.

The exact interface needs to be thought out, but here's a potential sketch:

```rust
struct AMP<Lo: Dtype, Hi: Dtype> { ... }

let q: Tensor<Rank1<5>, AMP<f16, f32>, _> = dev.zeros();
```

Things to figure out:
- How would certain layers choose an actual dtype from the two options in Amp? trait implementations?
- Is this even possible? Right now Dtype is just used directly for storage (ie. `Vec<Dtype>`), but this proposal is kind of a Dtype container. Does `Vec<AMP<f16, f32>>` even make sense?
- What are the generics on AMP? Are they even needed (i.e. implied f16/f32)?
- Need to write down exactly what operations are f16/f32, and what tensor/gradients are defaulted to (I think the lower precision).

coreylowman:
Resolves #424 

- [Breaking] Moves Unit/Dtype to new `dtypes` module
- Adds `dfdx::dtypes::AMP<F>`
- impl `Device<AMP<f16>>` for `Cpu`/Cuda`

Ops that cast up to f32 internally:
- sum_to
- adam
- sgd
- rmsprop

Other ops may be auto cast in the future.


nkoppel:
I'm curious: why `struct Amp<F>(F)` rather than `struct Amp(f16)`? Are there plans to support `Amp<f32>` as well? If not, the generic should be removed in my opinion.

src/dtypes/amp.rs
```diff
@@ -0,0 +1,571 @@
...
+    }
+}
+
+#[cfg(feature = "f16")]
+impl num_traits::AsPrimitive<AMP<half::f16>> for f64 {
+    fn as_(self) -> AMP<half::f16> {
+        AMP(half::f16::from_f64(self))
+    }
+}
+
```
nkoppel, line 187:
```suggestion
#[cfg(feature = "f16")]
impl<T: num_traits::AsPrimitive<E>, E> num_traits::AsPrimitive<AMP<E>> for T {
    fn as_(self) -> AMP<E> {
        AMP(self.as_())
    }
}
```

src/dtypes/amp.rs
```diff
@@ -0,0 +1,571 @@
...
+#[cfg(feature = "f16")]
+impl crate::shapes::SafeZeros for AMP<half::f16> {}
+
+#[cfg(feature = "f16")]
+impl crate::shapes::Unit for AMP<half::f16> {
+    const ONE: Self = AMP(half::f16::ONE);
+}
+
+#[cfg(feature = "f16")]
+impl crate::shapes::Dtype for AMP<half::f16> {}
```
nkoppel, line 212:
```suggestion
impl<T: crate::shapes::SafeZeros> crate::shapes::SafeZeros for AMP<T> {}

impl<T:  crate::shapes::Unit> crate::shapes::Unit for AMP<T> {
    const ONE: Self = AMP(T::ONE);
}

impl<T: crate::shapes::Dtype> crate::shapes::Dtype for AMP<T> {}
```

coreylowman:
> I'm curious: why `struct Amp<F>(F)` rather than `struct Amp(f16)`? Are there plans to support `Amp<f32>` as well? If not, the generic should be removed in my opinion.

@nkoppel I was thinking we could have `AMP<bf16>` as well, but I don't actually know if that is done in practice. If not, we can move to what you suggest. Can you check if anyone does that?

nkoppel:
PyTorch supports it, as seen [here](https://pytorch.org/docs/stable/amp.html), so it makes sense to leave room to do that. Might it make sense to use a type alias to refer to `AMP<f16>`  named something like `AMP16` to make this easier to use?

coreylowman:
👍 to type aliases

src/dtypes/amp.rs
coreylowman: Actually this leads to unconstrained type error on T

PR #813, opened 2023-07-13
accurate-gelu

### LINKED ISSUE #804: Support accurate GeLU as well as fast approximate GeLu
I was working on porting a GPT2 model to dfdx, and noticed a stubborn inconsistency in the output. It seems to arise from dfdx only supporting approximate GeLUs:
```rust
use dfdx::prelud::*;
let dev: Cpu = Cpu::default();
let data: Vec<f32> = vec![-5., -2., -1., 0., 1., 2., 5.];
let data: Tensor<Rank1<7>, f32, Cpu> = dev.tensor_from_vec(data, (Const::<7>,));
let activation: GeLU = GeLU::default();
println!("output: {:?}", activation.forward(data).as_vec())
```
yields:
```
output: [-2.9802322e-7, -0.04540229, -0.15880802, 0.0, 0.841192, 1.9545977, 4.9999995]
```

while pytorch yields
```python
to.nn.GELU()(to.Tensor([-5.,-2.,-1.,0.,1.,2.,5.]))
```
yields
```
tensor([-1.4901e-06, -4.5500e-02, -1.5866e-01,  0.0000e+00,  8.4134e-01,
         1.9545e+00,  5.0000e+00])
```

If instead we choose the approximate version
```python
to.nn.GELU(approximate='tanh')(to.Tensor([-5.,-2.,-1.,0.,1.,2.,5.]))
```
we get the same results as dfdx.
```
tensor([-2.9802e-07, -4.5402e-02, -1.5881e-01,  0.0000e+00,  8.4119e-01,
         1.9546e+00,  5.0000e+00])
```
piping to .tolist()
```
-2.980232238769531e-07,
 -0.04540228843688965,
 -0.15880799293518066,
 0.0,
 0.8411920070648193,
 1.9545977115631104,
 4.999999523162842
```

jcrist1:
This PR adds an accurate gelu function (used at least in in GPT2 model from huggingface) (see [Issue #804](https://github.com/coreylowman/dfdx/issues/804). Importantly in order to make the operators generic over  `Dtype`, I introduce an `Erf` trait that allows us to call `d_type.erf()` to get the error function of the value `d_type`. I currently am getting a compile error with feature `cuda`, and I have a hunch that this trait might be the issue. I'm having trouble debugging the build further as I don't have the cuda headers anywhere (working on a mac), so I would appreciate some help, so I was wondering if someone more familiar with the code might be able to point out the error.

For Github: Resolves #804


jcrist1:
Also I'm not at sure that the cuda actually supports the erf as I can't build cuda locally.

jcrist1:
Finally the naming is pretty terrible. Would it make sens to make a Gelu Activation an enum 
```rust
enum Gelu {
   Fast,
   Accurate
}
```
But then probably the `activation_impls!` macro couldn't be used

nkoppel:
Thank you for contributing! This looks good so far, I just have a few comments.
- To check the CUDA build without needing to call nvcc, use the command below. This won't be able to check your kernels, but should help you resolve your current CI errors.
```shell
cargo +nightly clippy -F cuda,ci-check
```
- I think it's best that we name the op/module 'AccurateGeLU' rather than 'GeLUCorrect' because it's easier to read and doesn't imply the the approximate GeLU op is incorrect.
- Be sure to document the difference between the accurate gelu op and the normal gelu.

src/tensor_ops/accurate_gelu/cpu_kernel.rs
```diff
@@ -0,0 +1,55 @@
...
+    fn erf(self) -> Self {
+        erf(self)
+    }
+}
+
+impl Erf for f32 {
+    fn erf(self) -> Self {
+        erff(self)
+    }
+}
```
jcrist1, line 28: I feel like these should be nearer to the Dtype trait, it feels a little odd putting them here, but I'm not that familiar with the code base so don't know the best place to put them yet.

jcrist1:
Okay @nkoppel I updated the name. Accurate GeLU is much better. I think distinguishing that the other GeLU is faster will help. I couldn't find an explicit citation for the fact that it's faster, but it seems to only require a single exponential, while the error function requires a much higher degree polynomial and still an exponential. I also beefed up the docs.

I feel like the changes to the docs for the activations don't fit in the code, but I'm guessing most people using the code will be using them as activations, rather than postfix operations so I added some info there even though it breaks the nice code block.

src/tensor_ops/accurate_gelu/cpu_kernel.rs
coreylowman: I actually like this location. Keeping things as close to where they are used is preferred IMO. Not sure if there's a standard practice for this though, but I like it 👍

src/tensor_ops/gelu/gelu.cu
coreylowman, line 1: Are the changes in this file just indentation? Can you revert them if so? Just for easier reviewing 😀

src/tensor_ops/accurate_gelu/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+///
+///
+/// Examples:
+/// ```rust
+/// # use dfdx::prelude::*;
+/// # let dev: Cpu = Default::default();
+/// let t = dev.tensor([-1.0, 0.0, 1.0, 2.0]);
+/// let r = t.accurate_gelu();
+/// ```
+pub fn accurate_gelu<S: Shape, E: Dtype, D: UnaryKernel<AccurateGeLUKernelOp, E>, T: Tape<E, D>>(
```
coreylowman, line 36: Thoughts on calling this `true_gelu`? I wonder if we should rename the other one to `approx_gelu`.

nkoppel: How should we deal with breaking name changes like this? A good approach may be to make aliases to the old naming marked as deprecated that direct users to the new names, at least for a major version or two.

nkoppel:
Resolves #804.

src/tensor_ops/accurate_gelu/mod.rs
coreylowman: Yeah that'd be a good way forward if we do want to rename it.

jcrist1: I have to admit I like `fast` vs. `accurate`, because `true` and `approx` gelu doesn't really describe the tradeoff: if there is a "true" gelu why would you _not_ choose it? but if there is an accurate and a fast gelu, then I know how to make a decision. What do y'all think? Otherwise fine with the approach of using an alias.

src/tensor_ops/gelu/gelu.cu
jcrist1: should be, and will do. LSP automatically updated them.

src/tensor_ops/accurate_gelu/mod.rs
jcrist1: I did it now with fast and accurate, but just to emphasize I'm not married to the naming. As repo owners tell me what's the best and I'll use that.

jcrist1: Didn't realise type aliased don't work for instantiating zero-size types, so I had to introduce a struct, but the implementation macro is not compatible with the `deprecated` macro, so I just implemented the traits by hand.

src/tensor_ops/fast_gelu/mod.rs
```diff
@@ -0,0 +1,79 @@
...
+
+    #[deprecated(since = "0.12.0", note = "Use `fast_gelu` instead")]
+    pub fn gelu(self) -> Self {
+        self.fast_gelu()
+    }
+
+    #[deprecated(since = "0.12.0", note = "Use `try_fast_gelu` instead")]
+    pub fn try_gelu(self) -> Result<Self, D::Err> {
+        self.try_fast_gelu()
+    }
```
nkoppel, line 57: Top two methods should link to `fast_gelu`, and deprecated items should have a link to their non-deprecated counterparts.

src/tensor_ops/fast_gelu/mod.rs
```diff
@@ -0,0 +1,79 @@
...
+
+#[allow(unused)]
+#[deprecated(since = "0.12.0", note = "use `FastGeLUKernelOp` instead")]
+pub type GeLUKernelOp = FastGeLUKernelOp;
+
+#[repr(C)]
+#[derive(Debug, Default, Copy, Clone)]
+pub struct FastGeLUKernelOp;
+
+/// [Fast Gaussian Linear Unit (GeLU)](https://paperswithcode.com/method/gelu). `0.5 * x * (1 + tanh(sqrt(2 / pi) * (x + 0.044715 * x^3)))`
```
nkoppel, line 17: These docs should include a link to AccurateGeLU

src/tensor_ops/accurate_gelu/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+/// 0.5 * x * (1.0 + erf(x / 2.0.sqrt()))
+/// ```
+/// As an accurate error function is [computationally expensive](https://en.wikipedia.org/wiki/Error_function#Numerical_approximations) it is
+/// possible to approximate the Gaussian Linear Unit with a hyperbolic tangent function `tanh`
+///
+/// ```text
+/// GeLU(x) ~ 0.5 ∗ x ∗ (1.0 + tanh((sqrt(2.0/π) ∗ (x + 0.044715 ∗ x^3)))
+/// ```
+///
+/// See [gelu](crate::tensor_ops::gelu::gelu) to use this approximation
```
nkoppel, line 26: This link needs to be fixed with the new naming

src/nn/activations.rs
```diff
@@ -38,6 +43,25 @@ activation_impls!(Abs, try_abs, #[doc="Calls [abs()]."]);
 activation_impls!(Softmax, try_softmax, #[doc="Calls [softmax()]."]);
 activation_impls!(LogSoftmax, try_log_softmax, #[doc="Calls [log_softmax()]."]);
 
+#[deprecated(since = "0.12.0", note = "please use `FastGeLU` instead")]
+#[derive(Default, Debug, Clone, Copy)]
+pub struct GeLU;
```
nkoppel, line 48: Needs to link to it's non-deprecated counterpart.

nkoppel (commented):
Just a few little fixes to documentation and this should be good to go! 

Note that I don't have write access to dfdx, I've just contributed a lot, so @coreylowman gets the final say on everything.

src/nn/activations.rs
jcrist1: link is 3 lines up

jcrist1:
@nkoppel alright should be addressed

coreylowman:
This looks good from me - any other updates planned here?

jcrist1:
I'm not planning on anything new.

coreylowman:
We can open another PR if there's something else to add/change!

Issue #810, opened 2023-07-12
Re-export `half::f16` in lib.rs if f16 feature is enabled


coreylowman:
This is so people don't have to add the half dependency themselves juts to get access to f16 type

Issue #809, opened 2023-07-12
Make Gradients::get_or_alloc_mut pub


coreylowman:
This will make it possible to implement something like nn::ZeroGrads outside of dfdx.

Related to #796

PR #814, opened 2023-07-13
Various small fixes

nkoppel:
A tiny pr that resolves multiple issues too small to deserve their own pr. It makes the following changes:

- Re-exports `half::f16` in `shapes/mod.rs`, resolving #810.
- Makes `Gradients::get_or_alloc_mut` public, resolving #809.
- Allows `get_or_alloc_mut` and `Gradients::get` to work with any implementer of TensorLike.
- Makes `Gradients::try_alloc_for` public
- Makes `Tensor::ghost` public (GhostTensor was already public, there just wasn't any way to construct it)
- Fixes multiple clippy warning, which may be new as of nightly rust 1.73.0


src/tensor_ops/matmul/cpu_kernel.rs
```diff
@@ -71,14 +71,14 @@ impl MatMulImpl<half::f16> for Cpu {
...
+                cp,
                 cstr[1] as isize,
                 cstr[0] as isize,
                 accum,
-                ap as *const gemm::f16,
+                ap,
                 astr[1] as isize,
                 astr[0] as isize,
-                bp as *const gemm::f16,
+                bp,
```
corey-lambda, line 81: We still need these because gemm might use a different version of f16. Internally gemm does a TypeId check against it's own version of f16, and the same type from two different versions of crates will have different TypeId

coreylowman, line 81: We still need these - internally gemm does a TypeId check against the type. So if gemm is using a different version of the half crate than dfdx, this will fail.

coreylowman (changes requested):
Looks good other than the gemm stuff

coreylowman (approved):
🎉

PR #831, opened 2023-07-26
Eric0000 refactor unstack

emchristiansen:
1. Adds an empty `.rustfmt.toml` so formatting for this project will be respected in workspaces with different formatting.
1. Adds `unstack`, which is the inverse of `stack`.

Note: I've never before made such liberal use of the Rust type system, so I basically did a lot of pattern-matching with `stack` to figure out what to do.
You'll probably want to review it carefully.
(The tests do pass, though.)

Note: I'm not confident in my ability to write correct CUDA kernels using your library, so I only implemented the CPU backend.

### DIFF
--- a/src/tensor/gradients.rs
+++ b/src/tensor/gradients.rs
@@ -171,6 +171,30 @@ impl<E, D: Storage<E>> Gradients<E, D> {
         let r_ref = unsafe { &*r_ptr };
         (l_refs, r_ref)
     }
+
+    #[inline]
+    pub(crate) fn many_mut_and_ref<L: Shape, R: Shape>(
+        &mut self,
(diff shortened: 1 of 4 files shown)


emchristiansen:
Hey, is this something you're interested in?
FYI I've been using this fn and it appears to work.

swfsql:
Nice! I'm interested in your unstack impl, and I hope to try it out soon. 

I'm actually interested in the opposite of a concatenate operation, eg. (6, 2) -> [(3, 2), (3, 2)], but after reading https://github.com/coreylowman/dfdx/issues/43#issuecomment-1153302228 I understood that it applies both ways. I could just first reshape (6, 2) -> (2, 3, 2) and then run unstack, getting [(3, 2), (3, 2)].

PR #849, opened 2023-08-13
Add complex numbers

### LINKED ISSUE #821: Add support for complex numbers
Pytorch has support for complex tensors and can also do Autograd for them.

https://pytorch.org/docs/stable/complex_numbers.html

https://pytorch.org/docs/stable/notes/autograd.html#autograd-for-complex-numbers

We can gate this behind a feature flag if we want. But it would be a nice to have.

opfromthestart:
Closes #821.
I had to use a proxy object to implement cuda traits. It build with `cargo build -F complex,cuda,nightly,f16`

### DIFF
--- a/src/dtypes/mod.rs
+++ b/src/dtypes/mod.rs
@@ -14,6 +14,111 @@ pub use amp::AMP;
 #[cfg(feature = "f16")]
 pub use half::f16;
 
+#[cfg(feature = "complex")]
+pub mod complex {
+    use core::ops::{Deref, DerefMut};
+
+    use cudarc::driver::DeviceRepr;
+    #[cfg(feature = "cuda")]
+    use cudarc::driver::ValidAsZeroBits;
+    use num_complex::Complex32;
+    use num_traits::{FromPrimitive, ToPrimitive};
+
+    #[derive(PartialEq, Debug, Default, Clone, Copy)]
+    pub struct Complex(Complex32);
+    impl Deref for Complex {
+        type Target = Complex32;
+
+        fn deref(&self) -> &Self::Target {
+            &self.0
+        }
+    }
+    impl DerefMut for Complex {
+        fn deref_mut(&mut self) -> &mut Self::Target {
+            &mut self.0
+        }
+    }
+    const fn c1() -> Complex {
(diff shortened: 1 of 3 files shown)


src/dtypes/mod.rs
```diff
@@ -14,6 +14,110 @@ pub use amp::AMP;
...
+pub mod complex {
+    use core::ops::{Deref, DerefMut};
+
+    #[cfg(feature = "cuda")]
+    use cudarc::driver::{DeviceRepr, ValidAsZeroBits};
+    use num_complex::Complex32;
+    use num_traits::{FromPrimitive, ToPrimitive};
+
+    #[derive(PartialEq, Debug, Default, Clone, Copy)]
+    pub struct Complex(Complex32);
```
coreylowman, line 27: Are we not able to use Complex32 directly?

coreylowman: Oh I see, it was for Cuda traits, missed that in description.

src/dtypes/mod.rs
```diff
@@ -14,6 +14,110 @@ pub use amp::AMP;
...
+    }
+    impl std::ops::DivAssign for Complex {
+        fn div_assign(&mut self, rhs: Self) {
+            self.0.div_assign(rhs.0)
+        }
+    }
+    #[cfg(feature = "cuda")]
+    unsafe impl ValidAsZeroBits for Complex {}
+    #[cfg(feature = "cuda")]
+    unsafe impl DeviceRepr for Complex {}
```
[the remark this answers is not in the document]
opfromthestart: I was unaware that you controlled cudarc. That would work yes.

opfromthestart:
I don't think it would work at all with cuda kernels. Not familiar with how cuda kernels work, can you overload operators for a custom complex struct?

coreylowman:
> I don't think it would work at all with cuda kernels atm. Not familiar with how cuda types work, can you overload operators for a custom complex struct?

We could potentially just merge this for Cpu and make sure we get compile errors for cuda.

You can overload operators in c++, but I'm not sure that's sufficient to make the kernels work in the same way as they do in rust land.

PR #860, opened 2023-08-28
feat: publish stacked API

### LINKED ISSUE #859: Expose AddDim to public API
I'm writing an API for my own library that has a generic Device as an input (instead of using `AutoDevice`, that way users can choose what device they want to pass in (or autodevice as default).

My implementation requires using the stacking util (using batches), which requires an extra generic restriction on some Shape: `AddDim<usize>`.

tristan-f-r:
Closes #859


src/tensor_ops/mod.rs
```diff
@@ -200,7 +200,7 @@ mod slice;
 mod softmax;
 mod sqrt;
 mod square;
-mod stack;
+pub(crate) mod stack;
```
coreylowman, line 203: `pub(crate)` makes it public only internal to the crate (e.g. dfdx).

I think you'd either want:
1. `pub mod stack` if we wanted to expose the whole stack module
2. `pub use stack::{TryStack, AddDim};` down below instead of exposing the whole module

coreylowman (changes requested):
Don't think this will actually help you, left comment with solutions 👍

PR #863, opened 2023-09-07
Adds 1d convolutions

### LINKED ISSUE #219: Support for conv1d
Currently `dfdx` supports `conv2d`, it would be if `conv1d` is supported out of the box. I know it'd be possible to use conv2d as a workaround but the code would not look clean.

coreylowman:
Resolves #219


coreylowman:
Thanks for the work on this @jcrist1!

PR #869, opened 2023-09-21
Sigmoid with residue

opfromthestart:
I don't know if this is something that should be in dfdx but it is useful for my use case where I need my model to return probabilities between 0 and 1, where each input is independent of all the others. If I use normal sigmoid I get vanishing gradients, so I made this.


opfromthestart:
right now it returns a minimum gradient of 0.0001, this maybe should be configurable.

coreylowman:
Since this can be accomplished by doing `(x.negate().exp() + 1.0).recip()` I'm inclined to not merge this. I know that this specialized kernel will be more efficient, but I think this is niche enough where I don't want to add to core.

Thanks for the PR!

opfromthestart:
This is not entirely right. While the forward pass is the same, the backwards pass is different. It makes sure that the gradients through the function are always at least some small epsilon. As in, if `x=100000000`, `sigmoid(x)=1`, which should make `sigmoid_dx(x)=0` but instead it is `0.00001`.

PR #873, opened 2023-10-12
Implementations for u8->u64 and i8->i64

timwedde:
This is my first Rust-related PR, so apologies for any inadvertent mistakes.

I picked up dfdx yesterday but in my application I'm dealing with u8's and u16's a lot, and I figured it would be quite handy if we could properly interface them with Numpy instead of having to cast to f32/64 for the data exchange.

This PR adds support for the following dtypes:
- `u8`
- `u16`
- `u32`
- `u64`
- `i8`
- `i16`
- `i32`
- `i64`

I wasn't sure what the policy of this project is in regards to testing, so I added two basic save/load tests for `u8` and `i8` to prove the concept. Let me know if I need to add all of the other tests as well. That'd be quite a few permutations in this case, so I didn't think it worth it for the initial PR.

### DIFF
--- a/dfdx-core/src/tensor/numpy.rs
+++ b/dfdx-core/src/tensor/numpy.rs
@@ -282,6 +282,166 @@ impl NumpyDtype for f64 {
     }
 }
 
+impl NumpyDtype for u8 {
+    const NUMPY_DTYPE_STR: &'static str = "u1";
+    fn read_endian<R: Read>(r: &mut R, endian: Endian) -> io::Result<Self> {
+        let mut bytes = [0; 1];
+        r.read_exact(&mut bytes)?;
+        Ok(match endian {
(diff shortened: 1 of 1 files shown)


ccaven:
For future reference, make sure to run `cargo fmt` so the automated testing doesn't give you a "Some checks were not successful" :)

timwedde:
Good tip, thank you, though in this case the failing check actually stems from clippy in a file I didn't even modify: `src/nn/add_into.rs:5:1`. Not sure why that is, but I guess it might be carry-over from some other commits on master before I forked and made this PR.
`cargo fmt` does nothing in my case, so I assume everything is in order, at least formatting-wise.

coreylowman:
@timwedde looks good to me, can you resolve conflicts and then I can force merge if pipelines don't pass

coreylowman:
Thanks! 🎉

PR #882, opened 2023-11-05
make device capable of stacking and concatenating usize type

optman:
which are necessary for preparing tensors for embedding.


coreylowman (approved):
👍

PR #886, opened 2023-11-08
Replace explicit features and paths on generated code

### LINKED ISSUE #885: Replace explicit features and paths on generated code
Some generated code has explicit feature detection and paths, such as `#[cfg(feature = "safetensors")]` and `::safetensors::Dtype`, but those would require the user app code (which depends on the dfdx lib) to have that feature, and to also directly depend on the safetensors lib.

An example:
https://github.com/coreylowman/dfdx/blob/4476b5ee19dc9cc446388545560c57e80cb086c8/dfdx-derives/src/lib.rs#L184-L192

For the feature detection, ideally it happens during code generation (but not be present in the generated code).  
For the `safetensors` dependency, it could be best to re-export it from the dfdx and then refer to it in the generated code. This problem also exists for when referencing `dfdx_core` and `dfdx_derives`.

Regarding the later, I'm not sure what's the best way forward. There are libs such as [proc-macro-crate](https://crates.io/crates/proc-macro-crate) that may help for the crate names that gets generated, but I believe there are other options such as allowing user-defined names like how [serde](https://github.com/serde-rs/serde/blob/4676abdc9e6bbbddfb33a00ce8d7e81e92f01120/serde_derive/src/internals/attr.rs#L533-L538) does.

swfsql:
Closes #885 .

- Disallow feature detection on generated code.
  - Move feature detection to before generating code.
  - Only emits the `SaveSafeTensors` and `LoadSafeTensors` derivations (as in, injecting the derivation command, not the final derivation itself) and the `#[serialize]` attr if the `dfdx/safetensors` feature is enabled.
  - Add `dfdx-derives/safetensors` feature.
  - Users apps no longer need to have a `safetensors` feature directly.
- Only use `::dfdx` as a path entrypoint on code generation
  - Re-export the `self` crate as `dfdx`.
    - See [related question](https://users.rust-lang.org/t/how-to-express-crate-path-in-procedural-macros/91274/10).
    - Does the same for `dfdx-core`.
  - Change the generated code so they always have the `::dfdx` path entrypoint.
    - User apps no longer need to depend on `dfdx-core` nor `dfdx-derives` directly.
  - Re-export `safetensors` lib.
    - User apps no longer need to depend on `safetensors` directly (as in, this is no longer required by the `SaveSafeTensors` and `LoadSafeTensors` derivations).

### DIFF
--- a/dfdx-core/src/lib.rs
+++ b/dfdx-core/src/lib.rs
@@ -110,6 +110,7 @@
 extern crate alloc;
 #[cfg(all(feature = "no-std", not(feature = "std")))]
 extern crate no_std_compat as std;
+extern crate self as dfdx_core;
 
 pub mod data;
 pub mod dtypes;
--- a/dfdx-derives/Cargo.toml
+++ b/dfdx-derives/Cargo.toml
@@ -16,3 +16,4 @@ dfdx-core = { path = "../dfdx-core" }
 
 [features]
 nightly = ["dfdx-core/nightly"]
+safetensors = ["dfdx-core/safetensors"]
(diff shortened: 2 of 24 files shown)


chelsea0x3b (approved):
Nice changes, thanks for this!

Issue #604, opened 2023-03-23
Consider WebGPU backend for inference in future


richardanaya:
Might be awesome to have inference in browser some day built with dfdx.

Cifram:
Also useful to have GPU acceleration that isn't specific to NVidia hardware, and WGPU should be able to provide that.

coreylowman:
It does seem like webgpu supports compute kernels - will need to look into this further. If this can support web stuff and enables AMD gpus, I'm inclined to do this one over #597.

Does anyone know if there are any GEMM/BLAS libraries for webgpu? Probably need to ask the same question for OpenCL

zojeda:
Perhaps this project can provide valuable insights: https://github.com/webonnx/wonnx

coreylowman:
Yes that is super helpful!!

kurnevsky:
It would be very nice to have it not only for inference but for training as well. There are people like me who don't like the quality of linux nvidia drivers and never buy their cards :)

jcrist1:
I'm enjoying working with the code, and working on a Mac, wgpu should give metals support for free so would love to help with this. I've also messed around with [wgpu](https://wgpu.rs/) before (but nothing serious).

antoineMoPa:
Looks like webgpu is at least partially supported now? Should we update the docs?

Issue #597, opened 2023-03-21
Add OpenCL device


coreylowman:
As brought up in reddit thread, a OpenCL device would be useful to support folks with AMD gpus.

Here are roughly the tasks that need to happen:
1. [ ] Add `opencl` feature flag
2. [ ] Create an `OpenCL` device struct under `src/tensor/opencl/device.rs` & add `impl DeviceStorage` for `OpenCL`
3. [ ] Implement tensor allocation methods under `src/tensor/opencl/allocate.rs`. See `src/tensor/cuda/allocate.rs` for examples
    1. [ ] impl ZerosTensor<E>
    2. [ ] impl OnesTensor<E>
    3. [ ] impl OneFillStorage<E>
    4. [ ] impl SampleTensor<E>
    5. [ ] impl CopySlice<E>
    6. [ ] impl TensorFromVec
    7. [ ] impl TensorToArray
4. [ ] Implement skeleton kernels for `OpenCL` under tensor_ops - tests should compile & run but fail
    1. [ ] Add module `opencl.rs` to all kernel folders
    2. [ ] Add a skeleton impl for all kernels (put `todo!()` in every kernel method)
    3. [ ] add `impl Device<E>` for `OpenCL` to `src/tensor_ops/utilities/device.rs`
5. [ ] Actually implement `OpenCL` kernels - this is highly parallelizable, each tensor op can be worked on separately. Every tensor op with a folder will need a kernel

jedbrown:
An alternative (lower effort and better supported by AMD) would be to provide HIP support. The interface is nearly perfect 1:1 with CUDA, including hipRTC.

Nabushika:
I've got some experience with OpenCL, especially in the context of generating kernels from compile-time representation of expressions, in C++. This sounds fun and I'd love to help! :)

jansol:
HIP would still leave out Intel GPUs though. (starting to become relevant now that they have discrete GPUs too!)

OpenCL would cover all vendors. Alternatively Intel seems to be pushing OneAPI as their "native" compute thing.

kstavro:
The first thing I asked when I joined the dfdx discord a couple of months ago was if OpenCL support was planned. Not an OpenCL expert, but I would like to be involved in this, if I can help.

jedbrown:
[...]

PR #893, opened 2023-11-27
Implement Webgpu device

favilo:
Tried an initial implementation for #604 

Based on the tasks from #597 
1. [x]  Add `webgpu` feature flag

2. [x]  Create an `Webgpu` device struct under `src/tensor/webgpu/device.rs` & add `impl DeviceStorage` for `WebGpu`
 
3. [x]  Implement tensor allocation methods under `src/tensor/webgpu/allocate.rs`. See `src/tensor/cuda/allocate.rs` for examples
   1. [x]  impl ZerosTensor
   2. [x]  impl OnesTensor
   3. [x]  impl OneFillStorage
   4. [x]  impl SampleTensor
   5. [x]  impl CopySlice
   6. [x]  impl TensorFromVec
   7. [x]  impl TensorToArray
 
4. [x]  Implement skeleton kernels for `Webgpu` under tensor_ops - tests should compile & run but fail
        
   1. [x]  Add module `webgpu.rs` to all kernel folders
   2. [x]  Add a skeleton impl for all kernels (put `todo!()` in every kernel method)
   3. [x]  add `impl Device<E>` for `Webgpu` to `src/tensor_ops/utilities/device.rs`
 
5. [ ]  Actually implement `OpenCL` kernels - this is highly parallelizable, each tensor op can be worked on separately. Every tensor op with a folder will need a kernel


coreylowman:
@favilo this is AWESOME!! I'm thinking we can merge this into main without actually implementing any of the kernels here, and then people can start adding them if they want. Does that sound good, or did you have plans to add more to this PR?

favilo:
I'm good with that, or you can make a new branch and we can add to that one. I figured we'd want to start adding kernels in other PRs mostly because I'm not entirely sure where to start with them. 

So I'm pretty much done with this particular code.

favilo:
Rebased onto main, again.

PR #895, opened 2023-12-02
Add `Prodigy` optimizer

### LINKED ISSUE #894: Add `Prodigy` optimizer
This is a request to add the [Prodigy](https://arxiv.org/pdf/2306.06101.pdf) optimizer - reference [pytorch implementation](https://github.com/konstmish/prodigy).

I think this could be helpful for newcomers as they presumably would have to bother less with manually setting optimizer configs.

swfsql:
Closes #894.

This is a tentative to implement this optimizer. It's based on a pytorch implementation from [here](https://github.com/konstmish/prodigy).  
That implementation is similar to adam, and so were the files where I started at. After the pytorch->dfdx translation, I gave it a try at cuda kernel - there's probably a lot of naivety and some incorrectness.  
For the fp16 side of things, I couldn't compile dfdx even without the prodigy changes, so I'm not sure if they are correct.

With the reasons above and some below, I'm leaving this as a draft.   
If I happen to find it good use in later experiments or find bugs in it, I hope to update it here.

#### Testing

I've added some very basic rust tests and compared with the equivalent pytorch test.

For every pytorch test, they are somewhat like this:

```python
import torch
import numpy as np
from prodigyopt import Prodigy

x = torch.tensor([[0.1, 0.2]], requires_grad=False).float()
m = torch.nn.Linear(2, 2, bias=False)
with torch.no_grad():
    w = torch.tensor([[3., 4.], [5., 6.]], requires_grad=True).float()
    m.weight = torch.nn.Parameter(w)
y = torch.tensor([[7e2, 8e2]], requires_grad=False).float()
loss_fn = torch.nn.MSELoss()
opt = Prodigy(m.parameters(), lr=1.) # this is the optimizer settings
preds = []
grads = []
weights = []
for i in range(0, 10):
    pred = m(x)
    preds.append(pred.detach().numpy().tolist())
    loss = loss_fn(pred, y)
    loss.backward()
    grads.append(m.weight.grad.detach().numpy().tolist())
    opt.step()
    weights.append(m.weight.detach().numpy().tolist())
    opt.zero_grad()
print(f"preds: {preds}")
print(f"grads: {grads}")
print(f"weights: {weights}")
```

Where the only change between each test happens on the `# this is the optimizer settings` line.

1. The first test compares against that default optimizer settings:
https://github.com/coreylowman/dfdx/blob/65cfd6a51805dda390c290e97312d67905a1e9dc/dfdx/src/nn/optim/prodigy.rs#L180-L182
2. The second compares with the settings:
https://github.com/coreylowman/dfdx/blob/65cfd6a51805dda390c290e97312d67905a1e9dc/dfdx/src/nn/optim/prodigy.rs#L225-L241
```python
opt = Prodigy(m.parameters(), lr=2e1, betas=(0.5, 0.25),beta3=0.4,eps=1e-8,use_bias_correction=True,safeguard_warmup=True,d0=1e-5,d_coef=0.5,growth_rate=1.02)
```
3. The third compares with the settings:
https://github.com/coreylowman/dfdx/blob/65cfd6a51805dda390c290e97312d67905a1e9dc/dfdx/src/nn/optim/prodigy.rs#L285-L295
```python
opt = Prodigy(m.parameters(), betas=(0.5, 0.25),beta3=0.4,weight_decay=1.0,decouple=False)
```
4. The fourth compares with the settings:
https://github.com/coreylowman/dfdx/blob/65cfd6a51805dda390c290e97312d67905a1e9dc/dfdx/src/nn/optim/prodigy.rs#L337-L347
```python
opt = Prodigy(m.parameters(), betas=(0.5, 0.25),beta3=0.4,weight_decay=1.0,decouple=True)
```

Besides this I've tried making a comparison with a unet experiment, and adam seemed to be much better, so I may have implemented something incorrectly. So this is another reason why this PR is still a draft.


swfsql:
I'll prioritize moving this experiment to a separate crate, but feel free to ping in case anyone have some question or suggestion.
Note that when I tried using this model I didn't had any success!

PR #899, opened 2023-12-05
Run tests with miri

### LINKED ISSUE #897: Run tests with miri
It's possible to use [miri](https://github.com/rust-lang/miri) to run some tests (cpu ones - so no `cuda` feature - and the ones that don't interact with python - so no `numpy` feature). Miri can be particularly useful to detect possible Undefined Behavior cases and memory leaks, although it can also have false positives (and false negatives - this depends on the tests coverage).

## Setup

The miri readme mentions using [cargo nextest](https://nexte.st) so tests are able to run in parallel, and for memory leaks to be detected.

```bash
# install miri
rustup +nightly component add miri

# install cargo nextest (Linux x86_64)
curl -LsSf https://get.nexte.st/latest/linux | tar zxf - -C ${CARGO_HOME:-~/.cargo}/bin
```

```bash
# cargo clean (run once when `cargo miri` is not used)
cargo clean

# run all tests (note: very slow and should take hours, assuming there are no errors)
MIRIFLAGS='-Zmiri-strict-provenance -Zmiri-symbolic-alignment-check -Zmiri-retag-fields -Zmiri-disable-isolation -Zmiri-tree-borrows' \
cargo miri nextest run --test-threads 1 --no-default-features --features std,fast-alloc,cpu,nightly,safetensors
```

Note that miri is very slow and takes an hour to run eg. the longest transformer test. So an accompanying issue (#898) would be to reduce the test sizes to their minimum.

### LINKED ISSUE #898: Reduce test sizes
The tests already are reasonably small and are fast to run eg. even on the CPU. But when simulating the execution through miri, various tests take more than half an hour, with one in transformers taking one hour. So one possibility to make this easier is to reduce the test sizes to their minimum, as long as they reach their testing goal.

Another benefit for a miri pass when reducing the tests is their memory requirements, as this tends to increase the longer the test needs to run, reaching 2GB of ram per test on the longest ones. So reducing their memory peak should make it easier to run more tests in parallel.

There are 367 tests, with many taking 10~20 seconds. So just focusing on what takes more than a few minutes could already be sufficient for the purpose of running all tests.

swfsql:
- Draft state
- Closes #897
- Closes #898

---

- Temporary patch on crossbeam.
  - Note that crossbeam is an indirect dependency and they are still to release a version that [passes](https://github.com/crossbeam-rs/crossbeam/pull/996) on this miri check. So to temporarily get this out of the way, you can patch crossbeam on the workspace `dfdx/Cargo.toml`.
  - [ ] Wait for crossbeam to make a new release and patch to that. 
  - [ ] Then try to update dependencies so the patch is no longer needed.
- [x] Check all test pass by some feature combinations.
  - [x] std,fast-alloc,cpu,nightly,safetensors
  - [x] std,cpu,nightly,safetensors
  - [x] no-std,fast-alloc,nightly
  - [x] no-std,nightly
- [x] Reduce test sizes so a miri pass is practical

---

To run a particular test, add it's rust path as a filter at the end of the miri run command:
```bash
# cargo clean (run once when `cargo miri` is not used)
cargo clean

# run all tests (note: very slow and should take hours, assuming there are no errors)
MIRIFLAGS='-Zmiri-strict-provenance -Zmiri-symbolic-alignment-check -Zmiri-retag-fields -Zmiri-disable-isolation -Zmiri-tree-borrows' \
cargo miri nextest run --test-threads 1 --no-default-features --features std,fast-alloc,cpu,nightly,safetensors
```

---

#### Miri UB errors
- None found, both under `dfdx` package (74 tests) and under `dfdx-core`(293 tests).

#### Memory/Thread leaks
- `nn::layers::add_into::tests::longer_network`
  - [x] Thread-leak error, disabling rayon on gemm for matmul seems to fix it. Commit: https://github.com/coreylowman/dfdx/pull/899/commits/c6e0642883cfeb61cda58283b96eb4505190607d
  - [ ] Investigate why the leak happened, try to re-enable rayon usage.
- `tensor::cpu::tests::test_reuse_allocation_on_clone_tensor`
  - [x] Memory-leak error, deallocating the cached allocations on `TensorCache` drop seems to fix it. Commit: https://github.com/coreylowman/dfdx/pull/899/commits/ae777d4699faa76eda4ad8167be959b6e0d9075a
  - [ ] Avoid panics during drops? Instead of panicking, can emit error message warning about possible memory leaks.
- None other found, both under `dfdx` package (74 tests) and under `dfdx-core`(293 tests).

#### Smaller tensors
Some tests have been reduced in tensor size so miri runs faster. Those changes are specific to the following commits:
- https://github.com/coreylowman/dfdx/pull/899/commits/9deca1bca27b2d4b1fe168d8779be53d1fcff9b6
- https://github.com/coreylowman/dfdx/pull/899/commits/86c9cbb92de74cd7026485c5eab5081a0bf70c9b

The tensor reduction is quite optional, they only makes it simpler to run miri "on everything". But if this change ends up persisting, this [wiki page](https://github.com/coreylowman/dfdx/wiki/Exporting-MultiHeadAttention-to-pytorch-for-unit-tests) should be updated. Finally, Idk how mha/transformer work, so I can't tell if the reduction shouldn't happen. That is, each test reduction should be reviewed to ack their reduction doesn't lose anything that should be tested.

---

Note that Miri depends on the test suite to find something, so having a bigger test suite may help. This is to say that a Miri check could be re-run in the future.
I haven't verified the cpu test coverage over the cpu source code, but a good way to go would be to have a high coverage (every relevant line of code getting "tested"). If more assurance would be desired, there could be more tests aiming at different characteristics of the source code, such as those related to branch coverage etc.


swfsql:
I'll be closing this as this was enough to indicate robustness on the library (at least in my opinion!), but nevertheless this can still be used as a reference whenever someone chooses to do a miri trial.

PR #912, opened 2024-01-26
Remove functions calling the deprecated `_MM_SET_FLUSH_ZERO_MODE`

rainiwu:
From https://github.com/rust-lang/stdarch/pull/1454, it seems that setting `_MM_SET_FLUSH_ZERO_MODE` results in "immediate Undefined Behavior", and as a result these functions were deprecated with 1.75.0. This PR removes the functions that include calls to `_MM_SET_FLUSH_ZERO_MODE`, fixing the warnings thrown when compiling on Rust versions 1.75.0 and newer.

### DIFF
--- a/dfdx-core/src/data/collate.rs
+++ b/dfdx-core/src/data/collate.rs
@@ -55,6 +55,7 @@ impl<A, B> Collate for Vec<(A, B)> {
 impl<'a, A, B> Collate for Vec<&'a (A, B)> {
     type Collated = (Vec<&'a A>, Vec<&'a B>);
     fn collated(self) -> Self::Collated {
+        #[allow(clippy::map_identity)]
         self.into_iter().map(|(a, b)| (a, b)).unzip()
     }
 }
--- a/dfdx-core/src/lib.rs
+++ b/dfdx-core/src/lib.rs
@@ -128,44 +128,6 @@ pub mod prelude {
     pub use crate::tensor_ops::*;
 }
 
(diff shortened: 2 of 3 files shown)


swfsql:
The other error (`error[E0277]: the trait bound cuda::device::Cuda: Conv1DKernel<f32> is not satisfied`) seems to be related to this:
https://github.com/coreylowman/dfdx/blob/4722a99d303f347d6088d95867d007c75ca6dd78/dfdx-core/src/tensor_ops/conv1d/mod.rs#L4-L5

I'm new to cuda/cudnn, but it appears that the cuda implementation of `Conv1DKernel` is behind that cfg condition.
I believe that to solve this, the same condition (`#[cfg(all(not(feature = "cudnn"), feature = "cuda"))]`) should be present on the Conv1DKernel Device bound, relaxing it for the cudnn case:
https://github.com/coreylowman/dfdx/blob/4722a99d303f347d6088d95867d007c75ca6dd78/dfdx-core/src/tensor_ops/utilities/device.rs#L116-L117

chelsea0x3b:
According to the new documentation of these functions, it seems like this functionality can be achieved with inline assembly? I'd much prefer that than entirely removing these functions.

swfsql:
I didn't read much material on this - but I'm afraid that every float operations would need to be implemented in asm (not sure)? if this is the case, I think this would require a lot of effort.

Could this be deferred to a later issue? Because if not, the CI would be blocking all prs until this is resolved.
From their discussion, they considered outright removing the functions (instead of deprecating it), so I think that leaving the code unchanged is not recommended.

PR #915, opened 2024-02-01
Add SiLU activation function

swfsql:
Adds [Sigmoid-Weighted Linear Unit (SiLU)](https://en.wikipedia.org/wiki/Rectifier_(neural_networks)) activation function. 
It's definition is `x * x.sigmoid()` and it's derivative is `x * sigmoid'(x) + sigmoid(x)`.

Note: I haven't made an actual pytorch test to compare, not even locally, but have made a bigger test that depend on this functionality (edit: in fact, inference only) and it appeared to work ok. So this PR should be considered a draft.


dfdx-core/src/tensor_ops/silu/cpu_kernel.rs
```diff
@@ -0,0 +1,20 @@
...
+    fn f(&self, x: &F) -> F {
+        *x / (F::one() + x.neg().exp())
+    }
+
+    // (1 + e^-x + x * e^-x) / (1 + e^-x)^2
+    // alternative: (e^x (1 + e^x + x)) / (1 + e^x)^2
+    #[inline(always)]
+    fn df(&self, x: &F) -> F {
+        let exp_nx = x.neg().exp();
+        F::one() + exp_nx + *x * exp_nx / (F::one() + exp_nx).powi(2)
```
ccaven: Are you missing parentheses here?

swfsql: Yes, thanks!
(I still haven't tested any backward op with SiLU myself, but hopefully it's correct now)

